PR #29, opened 2019-03-20
readme: mention the -encoding option

quasilyte:


YuriyNasretdinov:
Sorry, this option is left undocumented because the only other encoding that we support is `windows-1251` and PHP parser actually should not case about encoding at all because in PHP all strings are just bytes. This "feature" is not worth mentioning :)

PR #31, opened 2019-03-22
linter: add nameFilepath checker

quasilyte:
Checks that a name (namespace/interface/class/trait name)
correctly maps on a filesystem path.

That check should work even when running a linter for
a set of files with relative names since we translate
those into absolute paths during parsing.

To fix test that were failing due to the reason of "first.php"
filename, new helper functions filterReports and
getReportsSimpleFilename were added.

### DIFF
--- a/src/linter/lazy_test.go
+++ b/src/linter/lazy_test.go
@@ -13,6 +13,16 @@ import (
 
 var once sync.Once
 
+func filterReports(name string, reports []*Report) []*Report {
+	out := reports[:0]
+	for _, r := range reports {
+		if r.CheckName() == name {
+			out = append(out, r)
+		}
+	}
+	return out
+}
+
 func testParse(t *testing.T, filename string, contents string) (rootNode node.Node, w *RootWalker) {
 	once.Do(func() {
 		MaxFileSize = 10000
@@ -33,12 +43,16 @@ func testParse(t *testing.T, filename string, contents string) (rootNode node.No
 	return rootNode, w
 }
 
-func getReportsSimple(t *testing.T, contents string) []*Report {
+func getReportsSimpleFilename(t *testing.T, filename, contents string) []*Report {
 	meta.ResetInfo()
-	testParse(t, `first.php`, contents)
+	testParse(t, filename, contents)
 	meta.SetIndexingComplete(true)
-	_, w := testParse(t, `first.php`, contents)
+	_, w := testParse(t, filename, contents)
 	return w.GetReports()
+}
+
+func getReportsSimple(t *testing.T, contents string) []*Report {
+	return getReportsSimpleFilename(t, `first.php`, contents)
 }
 
 func TestLazy(t *testing.T) {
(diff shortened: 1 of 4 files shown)


src/linter/lazy_test.go
```diff
@@ -13,6 +13,16 @@ import (
 
 var once sync.Once
 
+func filterReports(name string, reports []*Report) []*Report {
+	out := reports[:0]
```
YuriyNasretdinov: This code modifies the original slice which is not documented and not expected here. I do not think we should be too stingy about extra memory allocations, especially in tests.

src/linter/root.go
```diff
@@ -921,6 +928,22 @@ func (d *RootWalker) parseFuncArgs(params []node.Node, parTypes phpDocParamsMap,
...
+	name, ok := ns.NamespaceName.(*name.Name)
+	if !ok {
+		return true
+	}
+
+	nsName := meta.NameToString(name)
+	pathSuffix := namespaceKey(nsName)
+	path := namespaceKey(filepath.Dir(d.filename))
+	if !strings.HasSuffix(path, pathSuffix) {
+		d.Report(ns, LevelDoNotReject, "nameFilepath", "namespace %s name doesn't match file system path", nsName)
```
YuriyNasretdinov: Actually there is no rule in PSR-4 that would require full namespace to match the path in the filesystem. E.g. "./acme-log-writer/lib/File_Writer.php" is a valid path for class "\Acme\Log\Writer\File_Writer"

I would suggest that we either:

1. Make this configurable (I am afraid there can be too many options available though)
2. Move this to a custom check

You can read more about PSR-4 at https://www.php-fig.org/psr/psr-4/ .

src/linter/root.go
```diff
@@ -1111,6 +1134,40 @@ func (d *RootWalker) enterConstList(lst *stmt.ConstList) bool {
...
+		}
+	}
+	for traitFullname := range d.meta.Traits {
+		if strings.HasSuffix(traitFullname, fileBaseName) {
+			badFilename = false
+		}
+	}
+
+	if badFilename {
+		// TODO: which node to use to report file-level issues?
```
YuriyNasretdinov: I suppose that we can use any class node or trait node. But we should save them somewhere then :).

src/linter/root.go
```diff
@@ -1134,4 +1191,16 @@ func (d *RootWalker) LeaveNode(n walker.Walkable) {
...
 	}
+
+	if n == d.rootNode {
+		d.leaveFile()
+	}
+}
+
+// namespaceKey returns ns with several transformations applied.
+func namespaceKey(ns string) string {
+	key := strings.Replace(ns, "\\", "/", -1)
```
YuriyNasretdinov: In Windows path separator is actually "\\", to I suggest you use `os.PathSeparator` here if you use this key to match the file path.

quasilyte:
Hmm.. Since this might be too strict for all project, maybe this check should be implemented as a custom check, outside of the main noverify distribution.

Issue #26, opened 2019-03-18
Failed parsing: If condition containing isset with variable variable expression


tonygoold:
When an if condition contains an isset that uses a "variable variable", `linter.(*BlockWalker).handleIf` panics because it expects a `*node.Identifier` for the variable name instead of a `*expr.Variable`.

## Code Example

```php
function get_global_var($name) {
    global $$name;
    if (isset($$name)) { // Panic occurs while parsing this line
        return $$name;
    }
    return null;
}
```

## Actual Behavior

```
Panic while parsing <filename>: interface conversion: node.Node is *expr.Variable, not *node.Identifier
```

## Expected Behavior

No panic.

YuriyNasretdinov:
Wow, didn't know that you can write isset this way :). Thank you for the report, I'll add some checks for that.

YuriyNasretdinov:
Should be fixed now.

PR #38, opened 2019-03-27
linter,meta: fix panic during isset($$x)

quasilyte:
Add support for variable-variable tracking inside a scope
as well, so we don't mark usages of $$x inside if body
as undefined. isset($$x) defines $x along $$x, so usages
of $x are permitted inside that body as well.
Doesn't work for arbitrary indirection level,
isset($$$x) is not supported (but it won't panic, as it did before).

Updates #26

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1270,9 +1270,23 @@ func (b *BlockWalker) handleIf(s *stmt.If) bool {
 		for _, isset := range a.issets {
 			for _, v := range isset.Variables {
 				if v, ok := v.(*expr.Variable); ok {
-					if !b.sc.HaveVar(v) {
-						b.addVar(v, meta.NewTypesMap("isset_$"+v.VarName.(*node.Identifier).Value), "isset", true)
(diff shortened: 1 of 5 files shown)


src/linter/block.go
```diff
@@ -1270,8 +1275,20 @@ func (b *BlockWalker) handleIf(s *stmt.If) bool {
...
+						b.addVar(v, meta.NewTypesMap("isset_$"+vn.Value), "isset", true)
+						defer b.sc.DelVar(v, "isset")
+					case *expr.Variable:
+						b.handleVariable(vn)
+						name, ok := vn.VarName.(*node.Identifier)
+						if !ok {
+							continue
+						}
+						b.addVar(v, meta.NewTypesMap("isset_$$"+name.Value), "isset", true)
 						defer b.sc.DelVar(v, "isset")
```
YuriyNasretdinov: There is no variable to delete here.

YuriyNasretdinov: Actually it is probably needed here. Everything is fine :)

PR #41, opened 2019-03-28
linter: add caseBreak check support fallthrough comment

### LINKED ISSUE #24: Allow to omit "break;" in switch statement if comment is provided
## Code Example

```php
<?php
switch ($something) {
case "one":
   $a++;
   // fallthrough
case "two":
   $a++;
   break;
}
```

## Actual Behavior

Case without break

## Expected Behavior

All ok with comment.

quasilyte:
Supported markers:
	"fallthrough",
	"fall through",
	"falls through",
	"no break",

Marker is expected to be a first comment line substring.
Both // and /* comment styles are handled.

Fixes #24

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -2,6 +2,7 @@ package linter
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/VKCOM/noverify/src/meta"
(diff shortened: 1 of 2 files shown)


src/linter/block.go
```diff
@@ -1450,7 +1451,11 @@ func (b *BlockWalker) handleSwitch(s *stmt.Switch) bool {
 
 		// allow to omit "break;" in the final statement
 		if idx != len(s.Cases)-1 && bCopy.exitFlags == 0 {
-			b.r.Report(c, LevelInformation, "caseBreak", "Case without break")
+			// allow the fallthrough if appropriate comment is present
+			nextCase := s.Cases[idx+1]
+			if !b.caseHasFallthroughComment(nextCase) {
+				b.r.Report(c, LevelInformation, "caseBreak", "Case without break")
```
YuriyNasretdinov: I suggest that we add some hint here, like this: "Case without break (you can add '// fallthrough' comment to mark that break is not needed here)"

quasilyte: Done.

PR #48, opened 2019-04-01
cmd: add -allow-checks argument

quasilyte:
Makes possible to run some checks in a separation,
which is useful in CI and while testing a new check.

Also makes it possible to have "disabled by default" checks
that are useful, but not for every linter user. It's also
easier to leave some checks disabled by default if they have
some correctness issues that can't be solved in a short period of time.

To figure out what checks are enabled by default we introduce
structured diagnostics info, linter.CheckInfo.
It's not mandatory to register a check with linter.DeclareCheck,
but it's the only way to associate metadata with diagnostics
implemented by any checker.

Non-declared checks can't be enabled by default, so
explicit -allow-checks entry should be specified for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,22 @@ $ noverify -exclude-checks arraySyntax,undefined -stubs-dir /path/to/stubs hello
 # No warnings
 ```
 
+### Using in CI / using explicit checks enable list
+
+For CI purposes it's usually more reliable to use an explicit list of checks to be executed,
+so updating a linter doesn't break your build only because new checks were added.
+
+The `-allow-checks` argument by default includes all stable checks, but you can override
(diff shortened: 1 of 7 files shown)


quasilyte:
@YuriyNasretdinov, what do you think?
It could make it possible to implement something like:
```sh
noverify doc checkName
```
And get diagnostic info (that is provided via `Comment` field).

quasilyte:
Ah, you probably mean the example code. Yeah, I agree, declaring code should be somewhere inside init function there. I'll fix that.

quasilyte:
Moved flag parsing inside cmd out of init function, so flags are bound after custom checks are declared.

PR #60, opened 2019-04-06
linter: add caseContinue check

quasilyte:
Detects "continue" (without level) usage inside switch
that behaves identically to "break".

As a consequence, it's either a bad style (if it's used
instead of break on purpose) or a bug (if "continue"
was intended, if there is a "for" loop outside).
Since both cases are bad, we report any continue (w/o level)
usage inside switch cases.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -20,6 +20,21 @@ import (
 	"github.com/z7zmey/php-parser/walker"
 )
 
+// loopKind describes current looping statement context.
+type loopKind int
+
+const (
+	// loopNone is "there is no enclosing loop" context.
+	loopNone loopKind = iota
+
+	// loopFor is for all usual loops in PHP, like for/foreach loops.
+	loopFor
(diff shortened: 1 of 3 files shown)


src/linter/block.go
```diff
@@ -1467,10 +1467,14 @@ func (b *BlockWalker) handleSwitch(s *stmt.Switch) bool {
...
-			if stmt != nil {
-				bCopy.addStatement(stmt)
-				stmt.Walk(bCopy)
+		for _, s := range list {
+			if s != nil {
+				bCopy.addStatement(s)
+				s.Walk(bCopy)
+				s, ok := s.(*stmt.Continue)
+				if ok && s.Expr == nil {
+					b.r.Report(s, LevelError, "caseContinue", "'continue' inside switch is 'break'")
```
YuriyNasretdinov: You can have continue inside if, for example. That should also count

PR #63, opened 2019-04-10
linter: add Filename() to the BlockContext

quasilyte:
Sometimes block-scoped checks need to know the filename as well.

Removed "full name" promise from the documentation as it may not
be the case if linter input was not created inside "cmd", which
does map paths to abs paths. In tests we don't do this, so
Filename() does not return the full name inside them.


YuriyNasretdinov:
Full filename is important for the internal VK checks. So I would rather ensure that it is always the case (tests don't count :))

Issue #66, opened 2019-04-15
No way to filter out issues by severity


sergey-shambir:
There are issues with low severity level (e.g. 'MAYBE').
Is there way to specify minimal severity level to filter issues with lower severity?

quasilyte:
There is no such feature right now. The severity level is something subjective and it's not very reliable.

Does explicit `-allow-checks` list work for you?

sergey-shambir:
Yes, `-allow-checks` works too.
Actually I'm working on patch which adds `-min-level` option support.
However, it's looks like current levels are not sequental. I'll look to my patch again tomorrow.

PR #68, opened 2019-04-17
Added option -min-level to filter issues by severity

sergey-shambir:
Also did refactoring:
- moved flags to flags.go (but global variables are evil)
- added new type SeverityLevel same as int
Syntax:
```bash
 -min-level=error|warning|info|hint|unused|maybe|syntax
```
Relates to issue #66

### DIFF
--- a/src/linter/custom.go
+++ b/src/linter/custom.go
@@ -85,7 +85,7 @@ type RootContext struct {
 // Report records linter warning of specified level.
 // chechName is a key that identifies the "checker" (diagnostic name) that found
 // issue being reported.
-func (ctx *RootContext) Report(n node.Node, level int, checkName, msg string, args ...interface{}) {
+func (ctx *RootContext) Report(n node.Node, level SeverityLevel, checkName, msg string, args ...interface{}) {
 	ctx.w.Report(n, level, checkName, msg, args...)
 }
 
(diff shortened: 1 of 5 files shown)


quasilyte:
I'll take a look today, thank.

quasilyte:
It would be easier to review if refactoring happened separately from the feature implementation.

src/linter/custom.go
```diff
@@ -176,14 +179,14 @@ var vscodeLevelMap = map[int]int{
...
-	LevelDoNotReject: "MAYBE  ",
-	LevelSyntax:      "SYNTAX ",
+var severityLevelNames = map[SeverityLevel]string{
+	LevelError:       "error",
+	LevelWarning:     "warning",
+	LevelInformation: "info",
+	LevelHint:        "hint",
+	LevelUnused:      "unused",
+	LevelDoNotReject: "maybe",
+	LevelSyntax:      "syntax",
```
quasilyte: I'm not sure about this severity level distribution. If you use min level filtering, you need a good sorting here and I don't know how to rank them from 1 to 7 properly. Are you sure you don't want to filter by using string list of prefixes?

sergey-shambir: I'm agree, it's better to pass enabled severity list.
However, it's inconvient to pass levels with command line flags.
So I suggest:
1) add config file support with separate PR
2) after that, add filtering by severity level names

quasilyte:
We should discuss and design the warning level/configuration separately.
I've applied your refactoring in #95.

quasilyte:
Discussion can be continued in #83.
Thanks.

PR #75, opened 2019-04-19
linter: don't tread "or die" as unconditional exit

quasilyte:
It's not correct to report code after "or die" as unreachable.
If it's desired to discourage die usages, it should be done
in a different way. "Unreachable code" error here is confusing.


quasilyte:
What about discarding exit flags effect of the "or" LHS walk?
This way, we cover all expressions that can perform exit, like `exit` itself and function calls that do call exit inside them.

YuriyNasretdinov:
Yes, that looks like it should work :).

PR #79, opened 2019-04-24
propagate exit flags from foreach/for/while/etc

### LINKED ISSUE #78: false positive for deadCode for blocks with no "return" (but with nested returns)
## Code Example

```php
function f($xs) {
    foreach ($xs as $v) {
        if ($v == 10) {
            return 10;
        }
    }
    exit; // Exits only condition inside a loop never fires
}

f([10]);
echo 123; // <- marked as unreachable
```

## Actual Behavior

`deadCode: Unreachable code` for `echo` statement.

## Expected Behavior

No warning.

quasilyte:
This makes deadCode properly handle "returns sometimes" case
for nested statements that create a nested block context.

Fixes #78


YuriyNasretdinov:
Thanks a lot for investigating and fixing this :)!

PR #80, opened 2019-04-29
make encoding handling customizable

quasilyte:
Make it possible for noverify extensions to specify
their own scheme for handling encodings.
Do this by providing file reader wrapping callback
that can, depending on the external logic,
either return original reader, or a new reader
that can handle input encoding by using internal
knowledge and read it as UTF-8 that linter expects.

This is especially useful in projects where more than
one encoding can be found inside project tree.

Changes overview:

- Move win-1251 encoding handling away from linter.
  It might go away in a next few releases, if that
  feature turns out to be unused by users.

- Add cmd.Main AfterFlagParse hook to simplify
  linter extension flags handling.
  Now user can bind new flags in main and
  use them inside AfterFlagParse function to
  finish initialization.

- Update /example. Use linttest package
  and demonstrate new cmd.Main hooks.

We also use io.ReadCloser instead of just Reader to make it
possible for custom wrappers of the reader to have some finalization
after reader is not needed anymore.

### DIFF
--- a/example/custom/custom_test.go
+++ b/example/custom/custom_test.go
@@ -1,47 +1,15 @@
 package main
 
 import (
-	"log"
-	"strings"
 	"testing"
 
-	"github.com/VKCOM/noverify/src/linter"
-	"github.com/VKCOM/noverify/src/meta"
-	"github.com/z7zmey/php-parser/node"
+	"github.com/VKCOM/noverify/src/linttest"
 )
 
-func init() {
-	linter.RegisterBlockChecker(func(ctx *linter.BlockContext) linter.BlockChecker { return &block{ctx: ctx} })
-	go linter.MemoryLimiterThread()
-}
-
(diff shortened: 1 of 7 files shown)


YuriyNasretdinov:
No, that's completely wrong! There should be no such thing as encoding in linter at all! There are two options:

1. Use binary-neutral parser like the one that PHP already has (which, sadly, php-parser isn't)
2. Claim that you only support UTF-8 :).

I do not see much profit for anyone except projects that have not yet moved to UTF-8. For these projects support of external linter is probably not a top priority...

Maybe I am wrong, but that's my point of view at least :).

YuriyNasretdinov:
And, as far as I remember, if you still want to improve encoding handling (which is a good intention at least :)) there still are some places that have specific code to handle encoding, e.g. in a language server. You can find them all by searching for "windows-1251" in the code :).

PR #85, opened 2019-04-30
accept unused-var-regex option

quasilyte:
Makes it possible to set "_" as a valid prefix for unused variables.

Useful for tuple-like returns that are destructured using list.

Only one value is used:
```
	list($_id, $_name, $email) = getUser(...);
```

If $_ is used everywhere, it's less opparent what fields are ignored:
```
	list($_, $_, $email) = getUser(...);
```

The next step is to warn on usages of variables that are matched
by IsDiscardVar. If such variable is read, it could be a mistake.

By default, only "$_" is permitted, as before.
For arbitrary patterns regexp is used.

### DIFF
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -57,6 +57,8 @@ var (
 	allowDisable      string
 	allowDisableRegex *regexp.Regexp
 
+	unusedVarPattern string
+
 	fullAnalysisFiles string
 
 	output string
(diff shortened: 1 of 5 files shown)


YuriyNasretdinov:
You do not have to write anything in list(...) if variable is unused:

list(,,$a) = array(1,2,3);

This change will also make it incompatible with other linters as only blank identifier ($_) is recognised as a valid name for an unused variable.

I was always against this change, but as far as I can see, E..r won in the end :).

quasilyte:
The good news is that by default it's `$_`.

PR #87, opened 2019-05-04
inherit base method param types

quasilyte:
If no other type info about params is available,
walk base types to find matching method and see whether
it has proper types info for that parameter.

Do this only for interfaces.

This makes it possible to resolve both input and output
types of methods that implement some interface. In turn,
this makes "undefined method" go away if method had no
PHPDoc that specified the parameter type that was used.

Phpstorm and phpdoc utility do more or less the same thing.

### DIFF
--- a/src/linter/cache.go
+++ b/src/linter/cache.go
@@ -16,7 +16,7 @@ import (
 	"github.com/VKCOM/noverify/src/meta"
 )
 
-const cacheVersion = 24
+const cacheVersion = 25
 
 var (
 	errWrongVersion = errors.New("Wrong cache version")
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -660,13 +660,48 @@ func (d *RootWalker) enterClassMethod(meth *stmt.ClassMethod) bool {
 		d.Report(meth.MethodName, LevelInformation, "phpdoc", "PHPDoc is incorrect: %s", err)
 	}
 
+	class := d.getClass()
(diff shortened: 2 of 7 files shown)


quasilyte:
@YuriyNasretdinov, maybe there are better ways?

YuriyNasretdinov:
The only real thing that bothers me is why is there a limit on params count :). It should be encoded as an argument and you just need to add something like wrap3 and unwrap3 functions

quasilyte:
Like using `<string>` of 1 character and check it's value? I though of adding `<uint8>` field later, but maybe storing it as a string is not a big deal.

YuriyNasretdinov:
You don't have to use string. You can use the code that formats length and
store it as an integer (shouldn't use binary encoding though)

quasilyte:
OK, I'll take a look.

quasilyte:
I'll try to reproduce it.

quasilyte:
We can get a pipe if we print wrapped type. It'll look like this:
```
	|0800FooClass0900barMethod
(First char is tab)
```
If we split type that are in this encoded representation, that will be a problem I guess.

YuriyNasretdinov:
I encountered some places where it is presumed that "|" is a separator between type names, although I am not sure where exactly it caused any problems. The only thing I remember is that it did mess things up :)

quasilyte:
Should be safe now. I'm using hex-encoding for `<uint8>`, so it consumes 2 bytes, but will never look like `|`. Tests also cover this particular peculiarity from now on.

PR #90, opened 2019-05-06
Handle late static binding

### LINKED ISSUE #89: Support `static` and `$this` types
## Code Example

```php
class Foo
{
    /** @return $this */
    public function bar()
    {
        return $this;
    }

    /** @return static */
    public function baz()
    {
        return new static();
    }
}
```

## Actual Behavior

For `static`, NoVerify behaves like it's `self`, which is not correct (since we need late static binding in those cases instead of an early one).

For `$this` it infers the "identity" type of `$this`, which is not useful. Whether it should map it into `self` or `static` is an open question though. I would stick to `static` in this case, but it could be more conservative to use `self` there, at least for now.

## Expected Behavior

`static` resolved into derived type on the use spots.
`$this` is either ignored since NoVerify can infer actual return types for `return $this` or mapped into some other type. It's also possible to handle `$this` in a special way, but it doesn't seem to yield any benefits over other forms (static/self).

See also https://github.com/yiisoft/yii2/issues/7045.

quasilyte:
This change enables more precise handling of `$this` and `static`
types that are quite common in Fluent APIs.

The idea is to carry the current class type along the type resolving
and map "static" pseudo-type into concrete class type.

Fixes #89

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1118,7 +1118,7 @@ func (b *BlockWalker) handleNew(e *expr.New) bool {
 
 	className, ok := solver.GetClassName(b.r.st, e.Class)
 	if !ok {
-		// perhaps something like 'new $class', cannot check this
+		// perhaps something like 'new $class', cannot check this.
 		return true
 	}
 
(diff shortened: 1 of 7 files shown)


quasilyte:
Sometimes `meta.GetClassName` is expected to return already resolved class name, for example, in `handleNew` we will get `undefined class "static"`, but we can't refine it into anything except `self` at that point. I'll fix that case and see other `meta.GetClassName` usages to figure out a better/general solution.

PR #91, opened 2019-05-07
Add tests for SimpleXMLElement/Iterator support

quasilyte:
Now that NoVerify understands `static` type,
given appropriate annotations, it can handle SimpleXMLElement
without any hacks.

An example of such (proper) types annotation:
	https://github.com/VKCOM/phpstorm-stubs/commit/faff1841c257873c35abfacf09cf8fac2fb3bdaa

### DIFF
--- a/src/linttest/exprtype_test.go
+++ b/src/linttest/exprtype_test.go
@@ -20,6 +20,29 @@ import (
 // TODO(quasilyte): better handling of an `empty_array` type.
 // Now it's resolved to `array` for expressions that have multiple empty_array.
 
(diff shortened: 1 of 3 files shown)


quasilyte:
I missed that case. Thanks.

quasilyte:
We have this info:
```php
    /**
     * Provides access to element's children
     * @param string $name child name
     * @return SimpleXMLElement
     */
    private function __get($name) {}
```
If we use `@return` info from `__get`, it should work.

quasilyte:
I'll work on #92 separately and will extend tests to include chaining.

PR #94, opened 2019-05-08
cmd: remove unused -git-local flag

quasilyte:
It's defined but not used.
Remove for now to avoid misliading info in -help output.

If it's needed, it can be returned back and
be actually implemented.

### DIFF
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -31,8 +31,7 @@ var (
 var (
 	outputFp io.Writer = os.Stderr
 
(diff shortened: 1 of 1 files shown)


YuriyNasretdinov:
Oh, it actually was (is?) implemented. It only analyzed changes that are not yet committed. Now I suppose it only requires work-tree argument to do the same thing.

quasilyte:
You are right, only `work-tree` is needed now. I think it's safe to remove that unused flag now.

PR #97, opened 2019-05-13
add discardVar check

quasilyte:
```
    add discardVar check
    
    Since $_ var usages are not tracked,
    warn if someone ends up using them.
    
    It's especially possible when discard vars are
    signified by $_ prefix, so $_id might be used.
    In that case, user should rename variable to
    $id or stop using it.
```


YuriyNasretdinov:
Why not merge this as well?

quasilyte:
This should be re-done with IR.

PR #98, opened 2019-05-14
improve handling of invalid phpdoc annotations

quasilyte:
When `@param` annotation misses variable type,
we should ignore it instead of using the variable name as a type.

Before this fix, NoVerify inferred \$a to be a type for $a
parameter out of `@param $a` line.

This fix is required to avoid false positives from arrayAccess.


YuriyNasretdinov:
As far as I know, you are not required to specify variable name in @param tag, so you can write the following and it will be recognised by most tools:

```
<?php
/**
 * @param int
 */
function test($a) {
// $a has type int here
```

I would suggest to only discard type information when it does not correspond to some regex or just ignore "types" that start with "$"

quasilyte:
Hmm, makes sense.
I'll send changes in a minute.

Issue #107, opened 2019-05-27
Parsing fails with right-to-left text


tonygoold:
## Code Example

```php
<? echo "اُردُو‎\n"; ?>
```

## Actual Behavior

The linter has a syntax error and is unable to examine the file.

```ERROR   syntax: Syntax error: syntax error: unexpected $unk at /path/to/utf8.php:2```

If the text is inside a class method, it can also show up as an "Empty root node" error.

## Expected Behavior

There should be no error.

## Notes

This appears to be a [php-parser issue](https://github.com/z7zmey/php-parser/issues/53) that was fixed, so the solution would be to upgrade the vendored php-parser library to the latest version.

quasilyte:
Can you use NoVerify from `dev-updated-parse` branch for a while and report if there are any new issues with it? We'll close this issue when that branch is merged into a master.

tonygoold:
No new issues so far, except for an unused import I accidentally left in (fixed in https://github.com/VKCOM/noverify/pull/117).

PR #108, opened 2019-05-27
Upgrade php-parser vendored dependency

tonygoold:
This fixes https://github.com/VKCOM/noverify/issues/107 by upgrading the vendored version of php-parser and making the necessary changes to use the new version.


quasilyte:
Thanks, I'll take a look at this week.
Need to make sure other code bases that depend on noverify don't break and all the tests are passing.

quasilyte:
Not all tests are passing. I'll try to figure out what is the case there.

quasilyte:
Seems like there is unpleasant peculiarity in this version of phpparser.
```
[1, 2] parsed as an array of 2 elements
[1, 2,] parsed as an array of 3 elements, last element being key=nil, val=nil
```
I'm not sure why it behaves this way. Probably deserves checking the issue tracker. And, maybe, bisecting the phpparser commits to see why/when it started to do it.

YuriyNasretdinov:
> I'm not sure why it behaves this way. Probably deserves checking the issue tracker. And, maybe, bisecting the phpparser commits to see why/when it started to do it.
That actually looks like a legit behaviour so I guess the logic needs to be fixed instead if it relies on the old behaviour somewhere.

quasilyte:
Maybe it is legit, but for static analysis, it's more trouble than benefit.
It could be a time to consider using some IR based or constructed out of AST, so we don't need those parser details leaking into the analyzer code. It will also make updating parser version easier (hopefullt, only transformation phase will need to be modified).

Just in case, I've filled https://github.com/z7zmey/php-parser/issues/90.

YuriyNasretdinov:
> Maybe it is legit, but for static analysis, it's more trouble than benefit.

I disagree. PHP 7.4 (I am not sure which exact version though) now supports short syntax for list(...):

instead of writing

```
list($a, $b, , $d) = [1, 2, 3, 4];
```

You could write

```
[$a, $b, , $d] = [1, 2, 3, 4];
```

Without information about skipped "array" entries there is no way of knowing which variables are assigned to which array elements.

YuriyNasretdinov:
Having some kind of intermediate respresentation could be useful anyway, of course. I thought about that a little but then I decided that it's not worth my time and gave up on that idea :).

quasilyte:
> [$a, $b, , $d] = [1, 2, 3, 4];

I see. I would make those LHS arrays a separate node kind that can have empty elements (that have meaning) while leaving normal arrays as they were because they can never have meaningful empty elements. For parser, it's not something very important, but this is where separate IR can also be handy in my opinion.

YuriyNasretdinov:
> I see. I would make those LHS arrays a separate node kind that can have empty elements (that have meaning) while leaving normal arrays as they were because they can never have meaningful empty elements. For parser, it's not something very important, but this is where separate IR can also be handy in my opinion.

While I don't disagree with your point, I would like to say that having empty elements can be useful from parser perspective, e.g. because it allows to reconstruct the source file a bit better if you are writing a code formatter, for example. It allows you to include all the details and it is up to you to decide how to use this information.

I am actually curious about how the IR is going to be implemented. I think that it can be done in a similar manner to PHP opcode generation, or you could actually use PHP opcode itself (PHP can generate it and save it to files starting with PHP7), but you will lose some features, e.g. you won't be able to check formatting or comments at all.

Also, one of the goals of php-parser project, if I remember correctly, was creation of CFG (Control Flow Graph) as well, which could help with static analysis a lot. It would be interesting to see where it goes :).

quasilyte:
My main complaint is that `[,,]` is not a valid PHP code if it's not used in that special LHS context.
Or maybe they were allowed globally in never version, I'm not sure. I'm not sure why they would allow them in RHS context, since `[1, , 2]` doesn't map to a meaningful value (but maybe relevant RFC covers that case).

YuriyNasretdinov:
You can omit the last element so that you can write arrays like this:

```
$a = [
  1,
  2,
];
```

this is equivalent to `$a = [1, 2];`.

As far as I understand, you cannot omit elements more than once, and you cannot omit elements anywhere but at the end.

I agree that parser should actually complain if someone tries to write `[1, , 2]` as use it as a value.

tonygoold:
@quasilyte I managed to get the tests passing. Because freefloating text (e.g., comments) and position have been moved into the various structs that implement `node.Node`, it broke how types are associated with some nodes. I've added a function for comparing nodes that works around this and confirmed tests are passing now: https://github.com/VKCOM/noverify/pull/108/commits/348e927a31508f33e313b09400db2c0dfaf65569#diff-6f02aeb03093234744c129907fb4999bR388

quasilyte:
Let's merge it in a dedicated branch, try it out for some time and if there are no problems detected, we can integrate it into master. Some external things depend on the master, so I can't make this transition right now. 

I would like to get your feedback on using updated noverify, whether anything gone wrong with it, for example.

PR #116, opened 2019-06-11
cmd: add -output-json option

### LINKED ISSUE #115: Add -output-json flag
If we want to process linter results, it can be easier if we had JSON or another kind of machine-readable format.

I wanted to start from the "short" format that doesn't include the context line, but since it still would require more processing that JSON, we could start from JSON right away.

To avoid the temptation to extend the supported format list, we can stick to `-output-json` bool flag instead of giving `-output-format` string argument. If other transformations are needed, they can be done over JSON like `noverify -output-json args... | transform-json > result.txt`.

quasilyte:
With -output-json linter writes a single JSON object with
reports to the output (stderr or file).

This JSON can be read by external program.

For Go external programs it's required to either use
map[string]interface{} or express the format manually by
something like:

	type reportsList struct {
		Reports []struct {
			CheckName string `json:"check_name"`
			Severity  string `json:"severity"`
			Context   string `json:"context"`
			Message   string `json:"message"`
			Filename  string `json:"filename"`
			Line      int    `json:"line"`
			StartChar int    `json:"start_char"`
			EndChar   int    `json:"end_char"`
		}
	}

Fixes #115


src/cmd/main.go
```diff
@@ -205,6 +206,7 @@ func buildCheckMappings() {
 }
 
 func analyzeReports(diff []*linter.Report) (criticalReports int) {
+	filtered := diff[:0]
```
YuriyNasretdinov: To be honest, I am not a huge fan of micro-optimisations like this. analyzeReports probably is not a bottleneck so you are free to just allocate a new slice with enough capacity, for example. Code like this is harder to verify, at least for me, and it does not provide too much benefits in terms of performance in this case.

src/linter/root.go
```diff
@@ -83,6 +84,34 @@ func (r *Report) CheckName() string {
 	return r.checkName
 }
 
+// MarshalJSON is used to write report in its JSON representation.
+//
+// Used for -output-json option.
+func (r *Report) MarshalJSON() ([]byte, error) {
+	type jsonReport struct {
+		CheckName string `json:"check_name"`
+		Severity  string `json:"severity"`
```
YuriyNasretdinov: I think that it would be simpler to provide MarshalJSON method for Severity field and make it a proper type as well.

quasilyte:
I'll send follow-up PR to fix review comments.

PR #119, opened 2019-06-14
fix false positives with instanceof

quasilyte:
When instanceof is inside if condition, it introduces
a new tracked var via b.addVar().
It seems to be incorrect for cases where added variable
was already used above.
Given that we enter (walk) instanceof LHS before
we add that new variable, instanceof itself doesn't
end up marking that variable as used.

This change makes linter stop tracking instanceof LHS.
(We add it to the scope directly instead.)


YuriyNasretdinov:
The bug is pretty unintuitive :). I wonder how much time you spent on it. I think that the fix is a bit dirty but it's ok I guess.

PR #122, opened 2019-06-20
Write "you can't disable linter for file" to stderr

quasilyte:
If we write those to outputFp, we destroy JSON output.
Instead of doing jsonOutput branching and using different
output descriptors for it, just write that warning to
stderr in all cases.


YuriyNasretdinov:
You might also want to include this information to linter JSON output as it makes it more obvious to the user why linter disable doesn't work for the specific file

PR #131, opened 2019-07-10
Don't copy block walker and custom walkers for nested blocks

quasilyte:
Since custom block walkers can't reliably propagate their state
to the nested custom block checkers (that are created by constructor
function), they loose needed state when entering these blocks.

Since PHP is function-scoped and all other blocks do not create
real scopes, it's intuitive that function block walker lives for
the entire function body. This is especially important if that
checker collects function-bound info, like global'ed variables.

Moved all non-shared (between contexts) data into a separate struct
that became a part of BlockWalker. All other fields are considered
to be safe to share between nested contexts.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -65,8 +65,9 @@ func FlagsToString(f int) string {
 
 // BlockWalker is used to process function/method contents.
 type BlockWalker struct {
-	sc *meta.Scope
-	r  *RootWalker
+	ctx *blockContext
+
+	r *RootWalker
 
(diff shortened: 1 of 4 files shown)


quasilyte:
The major problem was the inability to reliably pass "parent" custom block checker state to the custom child custom block checker.

But if we look at it at a different angle, it seems that avoiding a problem of passing the partial block state around can be mitigated if we don't clone custom block checkers at all.

YuriyNasretdinov:
LGTM

Issue #124, opened 2019-06-21
Add @property support


quasilyte:
Handle `@property` annotations.

See #123.

quasilyte:
Done.

Issue #123, opened 2019-06-21
False positive: Call to undefined method and @property


kratorman:
## Code Example

```php
/**
 * @property Service $service
 */
class Container {
    private $m = [];
    
    public function __get($name)
    {
        return $this->m[$name];
    }
}

class Service {
    public function action() {}
}

$c = new Container();
$c->service->action();
```

## Actual Behavior
```
ERROR   undefined: Call to undefined method {}->action() at /test.php:20
$c->service->action();
             ^^^^^^
```

## Expected Behavior

There should be no error cause type defined in PHPDoc.

YuriyNasretdinov:
Yeah, currently there is no support for defining methods and properties in phpdoc. Shouldn't be hard to add though

quasilyte:
I think it should be fixed as `@property` annotation is now supported.

PR #132, opened 2019-07-12
basic @property annotation support

quasilyte:
Register @property-annotated names as instance properties.

Updates #123 #124


src/linter/root.go
```diff
@@ -810,6 +821,70 @@ func (d *RootWalker) enterClassMethod(meth *stmt.ClassMethod) bool {
...
+	var result classPhpDocParseResult
+	var phpDocErrors []string
+
+	if doc == "" {
+		return result, nil
+	}
+
+	result.properties = make(meta.PropertiesMap)
+
+	// TODO(quasilyte): reduce code duplication in various phpdoc parsing methods.
```
YuriyNasretdinov: I guess you could do it in this pull request as it should not be that hard and I think that I have started to do it already but I am not 100% sure about that.

YuriyNasretdinov: There exists phpdoc package and phpdoc.Parse() method. I am not sure why I didn't use it everywhere

src/linter/root.go
```diff
@@ -1038,75 +1003,57 @@ func (d *RootWalker) parsePHPDoc(doc string, actualParams []node.Node) (phpDocPa
...
-		ln = strings.TrimPrefix(ln, "*")
-		ln = strings.TrimSuffix(ln, "*/")
-
-		if strings.Contains(ln, "@deprecated") {
-			parts := strings.Split(ln, "@deprecated")
+	for _, part := range phpdoc.Parse(doc) {
+		if part.Name == "deprecated" {
 			result.info.Deprecated = true
-			result.info.DeprecationNote = strings.TrimSpace(parts[1])
+			// strings.Join here is unfortunate.
```
YuriyNasretdinov: I guess we could retain the original string somewhere as well to avoid having to join back the string

PR #134, opened 2019-07-15
use embedded stubs by default

quasilyte:
Use github.com/go-bindata/go-bindata to pack github.com/VKCOM/phpstorm-stubs
into src/cmd/stubs and use it from the compiled binary from default.

To re-generate stubs, the following steps should be used:

	1. Clone github.com/VKCOM/phpstorm-stubs into src/cmd/stubs
	2. Run "go generate"

(Comments inside code tell the same.)

The phpstorm-stubs folder (git repo) is ignored,
generated file (phpstorm_stubs.go) is commited.

If user provides explicit -stubs-dir override, use that instead.

### DIFF
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -329,7 +329,8 @@ func loadEmbeddedStubs() error {
 	linter.ParseFilenames(readStubs)
 	meta.Info.InitStubs()
 
-	if errorsCount != 0 {
+	// Using atomic here for consistency.
+	if atomic.LoadInt64(&errorsCount) != 0 {
 		return fmt.Errorf("failed to load %d embedded files", errorsCount)
 	}
 
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -13,11 +13,19 @@ import (
 	"runtime/pprof"
 	"strings"
 
+	"github.com/VKCOM/noverify/src/cmd/stubs"
 	"github.com/VKCOM/noverify/src/langsrv"
 	"github.com/VKCOM/noverify/src/lintdebug"
 	"github.com/VKCOM/noverify/src/linter"
 	"github.com/VKCOM/noverify/src/meta"
 )
+
+// Line below implies that we have `https://github.com/VKCOM/phpstorm-stubs.git` cloned
+// to the `./src/cmd/stubs/phpstorm-stubs`.
+//
+// Since there is no -include flag, we use -ignore to filter-out all non-php files.
+//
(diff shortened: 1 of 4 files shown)


src/cmd/main.go
```diff
@@ -278,3 +289,50 @@ func setDiscardVarPredicate() error {
...
+				continue
+			}
+			ch <- linter.FileInfo{
+				Filename: filename,
+				Contents: data,
+			}
+		}
+	}
+
+	if errorsCount != 0 {
```
YuriyNasretdinov: There probably is a "race" here as the whole point why readStubs writes to a channel is that it is executed in a separate goroutine. You can try to run noverify with "-race" option to ensure that it is in fact a race condition here. You need to use atomic increment here probably.

src/cmd/main.go
```diff
@@ -13,12 +13,20 @@ import (
...
 	"github.com/VKCOM/noverify/src/linter"
 	"github.com/VKCOM/noverify/src/meta"
 )
 
+// Line below implies that we have `https://github.com/VKCOM/phpstorm-stubs.git` cloned
+// to the `./src/cmd/stubs/phpstorm-stubs`.
+//
+// Since there is no -include flag, we use -ignore to filter-out all non-php files.
+//
+//go:generate go-bindata -pkg stubs -nometadata -o ./stubs/phpstorm_stubs.go -ignore=.*\.(.|..|[^p][^h][^p]|.{4,})$ ./stubs/phpstorm-stubs/...
```
YuriyNasretdinov: Are you 100% sure there is no better way to do this?

YuriyNasretdinov: Also this regex does not seem to match `meta/.phpstorm.meta.php` file which is really important.

quasilyte: >  Are you 100% sure there is no better way to do this?

Which part is implied here?

- go-generate?
- go-bindata?
- or regexp?

I'll try improving regexp. There is no opposite of "ignore" in bindata. The only alternative I see right now is to remove redundant files from the repo or pass results of "find" (or similar utility) as inputs to bindata (although I don't remember whether go generate permits sub-processes).

src/cmd/main.go
quasilyte: And I actually put check before parsing call.
Thanks for spotting that. I'll fix that (and race things as well).

src/cmd/main.go
YuriyNasretdinov: I was talking about go-bindata call itself. Maybe it there is a way to exclude .git

quasilyte: Yep, excluding `.git` is trivial. It'll include some redundant files, but I think it's not a problem.
I was aiming to include only php files from start, so the generated file has minimal size.

src/cmd/main.go
YuriyNasretdinov: You need to read using atomic instructions as well :).

PR #138, opened 2019-07-22
add keywordCase checker

### LINKED ISSUE #136: Require lowercase keywords
## Code Example

```php
CONST MY_CONST = 1;
```

## Actual Behavior

No warning.

## Expected Behavior

Linter should suggest to use `const` there (instead of `CONST`).

See <https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull>:

> PHP keywords MUST be in lower case.

quasilyte:
Handles keywords listed in https://php.net/manual/en/reserved.keywords.php
expect function-like array, isset and alike.

This check implements PSR-2 rules about keywords casing.
See https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull

For matching tokens we use poor man's tokenization on the input sources,
since parsed AST nodes do not contain which literal text was used
inside input to represent a given node.

We only tokenize a single string: the first or the last line
of the statement. Last line is needed to match trailing keywords as
"endif" and "endfor".

Fixes #136

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -230,7 +230,19 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
 		b.checkRedundantCast(s.Expr, "string")
 	case *cast.Array:
 		b.checkRedundantCast(s.Expr, "array")
+	case *stmt.Goto:
+		b.r.handleKeyword(n, "goto")
+	case *stmt.Throw:
+		b.r.handleKeyword(n, "throw")
+	case *expr.InstanceOf:
+		b.r.handleKeyword(n, "instanceof")
+	case *expr.Yield:
+		b.r.handleKeyword(n, "yield")
+	case *expr.YieldFrom:
+		b.r.handleKeyword(n, "yield")
+		b.r.handleKeyword(n, "from")
 	case *stmt.Global:
+		b.r.handleKeyword(n, "global")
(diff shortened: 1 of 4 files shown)


quasilyte:
**sigh** it still has some problems.
For example, string literals.

quasilyte:
I might be missing even more cases. Extending the function to parse more can be a bad road to take. But right now I see no real alternative.

YuriyNasretdinov:
To be honest this (the suggested implementation) is probably one of the reasons why I didn't consider adding any formatting features to the linter. My opinion is that we should only do analysis that AST allows us to do relatively easily.

And you definitely should not try to tokenise the file yourself as php-parser already includes tokeniser and it builds AST on top of tokens that are available to it.

I think that you need to add missing features to `php-parser` first and only then try to do anything like that. I suppose that some features that you need are already present in newer versions of `php-parser` too.

quasilyte:
The problem is that a single line like
```php
CLASS Foo {
```
Is not complete and can't be tokenized by the standard parser.
AFAIK, tokenization results are not available after parser finishes its job.

It would be nice if parsed nodes retained their keywords info in some way.

YuriyNasretdinov:
Yes, this is what I meant: it is parser that needs to provide that information, it should not be us trying to re-tokenise some parts of the source code, especially line-by-line.

PR #149, opened 2019-08-13
Associate absolute paths with git files

quasilyte:
When running NoVerify in normal (non-git) mode,
all filenames are turned into full filenames (abs paths).
It's not the case with git mode, since most code paths
recorded relative names.

Evaluate .git-containing folder once, during linter init
and use it to get absolute path later, when collecting
filenames provided by git catter.


quasilyte:
Should be done from scratch now.

PR #155, opened 2019-08-19
Recognize variable uses in property name subexpressions

tonygoold:
This fixes false positives saying that `$x` is unused when it is used as a property (or static property) name, e.g., `->$x` or `::$$x`.


tonygoold:
I think the names stayed the same, but I changed the order of some of the code, so the diff makes it look like more changes than there really were.

YuriyNasretdinov:
Yes, sorry, you're right. I just got confused, never mind the change in naming :).

quasilyte:
Sorry for the delay and thanks for contribution!

PR #175, opened 2019-09-25
resolve elemOf for ArrayAccess and Traversable

quasilyte:
For ArrayAccess use return type of offsetGet() method.
For Traversable (Iterator, etc.) use return type of current() method.

This behavior is in sync with what PhpStorm does.

### DIFF
--- a/src/solver/solver.go
+++ b/src/solver/solver.go
@@ -24,7 +24,7 @@ type resolver struct {
 }
 
 func (r *resolver) resolveType(class, typ string) map[string]struct{} {
-	res := r.resolveTypeNoSpecialization(class, typ)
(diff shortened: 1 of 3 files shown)


src/solver/solver.go
```diff
@@ -24,6 +24,17 @@ type resolver struct {
...
+
+	if _, ok := res["static"]; ok {
+		delete(res, "static")
+		res[class] = struct{}{}
+	}
+
+	return res
+}
+
+func (r *resolver) resolveTypeNoSpecialization(class, typ string) map[string]struct{} {
```
YuriyNasretdinov: Late static binding :)

PR #177, opened 2019-09-26
handle __construct() method call in argCount diagnostic

### LINKED ISSUE #176: argCount: check for constructor arguments count
## Code Example

```php
<?php

class Foo {
  public function __construct($x) {
  }
}

function f() {
  $_ = new Foo();
  $_ = new Foo;
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Warnings about "not enough arguments".

quasilyte:
Do it by analyzing new expressions.

If instantiated object class doesn't define __construct method,
check for a method that have the same name as a type name.

Fixes #176

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1183,20 +1183,14 @@ func (b *BlockWalker) handleNew(e *expr.New) bool {
 		return true
 	}
 
-	info, ok := meta.Info.GetClass(className)
-	if !ok {
+	if _, ok := meta.Info.GetClass(className); !ok {
 		b.r.Report(e.Class, LevelError, "undefined", "Class not found %s", className)
 	}
 
 	// Check implicitly invoked constructor method arguments count.
-	ctor, ok := info.Methods["__construct"]
+	ctor, _, ok := solver.FindMethod(className, "__construct")
 	if !ok {
-		// TODO(quasilyte): we might want to catch case-insensitivity issues.
-		// Class may be called "Foo", but "foo" method can be its constructor.
-
-		// If constructor method uses className name instead of __construct.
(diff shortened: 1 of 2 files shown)


src/linter/block.go
```diff
@@ -1176,10 +1183,31 @@ func (b *BlockWalker) handleNew(e *expr.New) bool {
...
 	}
 
-	if _, ok := meta.Info.GetClass(className); !ok {
+	info, ok := meta.Info.GetClass(className)
+	if !ok {
 		b.r.Report(e.Class, LevelError, "undefined", "Class not found %s", className)
 	}
 
+	// Check implicitly invoked constructor method arguments count.
+	ctor, ok := info.Methods["__construct"]
```
YuriyNasretdinov: Constructor can be defined in children so that's why I use solver.FindMethod everywhere. Constructors can also be private (or protected) so it be also worth trying to check the context as well.

src/linter/block.go
```diff
@@ -1176,10 +1183,31 @@ func (b *BlockWalker) handleNew(e *expr.New) bool {
...
+	if !ok {
+		// TODO(quasilyte): we might want to catch case-insensitivity issues.
+		// Class may be called "Foo", but "foo" method can be its constructor.
+
+		// If constructor method uses className name instead of __construct.
```
YuriyNasretdinov: I think it is no longer supported in PHP 7.x so I would not even try to check it.

quasilyte:
Fixed.
I'll check for old-style constructors in a separate check.

quasilyte:
@YuriyNasretdinov thank you for the review!

PR #180, opened 2019-10-01
Update expected value in phpdoc parser test

tonygoold:
The Line and ParamsText fields were added without updating the expected value,
resulting in the test failing.


quasilyte:
We really need to integrate pre-merge check already..

PR #194, opened 2019-10-03
Remove unused method and const #192

### LINKED ISSUE #192: Remove unused method
We're about to enable Go linter checks on CI, as a part of that, I would like to see all issues inside current codebase fixed (at least ones that are easy to fix).

This issue is about warnings found by go staticcheck linter.

```
src/linter/block.go:128:23: func `(*BlockWalker).copyContext` is unused (unused)
func (b *BlockWalker) copyContext() *blockContext {
                      ^
```

Unused method should be removed.

u5surf:
Fixes #192 

I removed unused method and const following golangci-lint said
```
$ golangci-lint run --disable-all -E unused
src/langsrv/langsrv.go:74:38: U1000: func `IMessage` is unused (unused)
func writeMessage(message interface{ IMessage() }) error {
                                     ^
src/linter/block.go:28:2: U1000: const `loopNone` is unused (unused)
	loopNone loopKind = iota
	^
src/linter/block.go:128:23: U1000: func `(*BlockWalker).copyContext` is unused (unused)
func (b *BlockWalker) copyContext() *blockContext {
                      ^
```

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -125,10 +125,6 @@ func (b *BlockWalker) addStatement(n node.Node) {
 	b.statements[assignment] = struct{}{}
 }
 
-func (b *BlockWalker) copyContext() *blockContext {
-	return copyBlockContext(b.ctx)
-}
-
 func (b *BlockWalker) reportDeadCode(n node.Node) {
 	if b.ctx.deadCodeReported {
 		return


quasilyte:
Thanks!
Can we keep `loopNone` for a while?

quasilyte:
I also believe interface parameter used a method declaration to avoid passing "any object" there, even if it doesn't use defined method inside.

Issue #187, opened 2019-10-03
Fix redundant conversions 


quasilyte:
We're about to enable Go linter checks on CI, as a part of that, I would like to see all issues inside current codebase fixed (at least ones that are easy to fix).

This issue is about warnings found by go unconvert linter.

```
src/linter/report.go:213:27: unnecessary conversion (unconvert)
	for i, ch := range string(r.startLn) {
	                         ^
src/langsrv/refs.go:188:16: unnecessary conversion (unconvert)
	endLine := int(pos.EndLine) - 1
	              ^
src/langsrv/refs.go:196:36: unnecessary conversion (unconvert)
			Start: vscode.Position{Line: int(pos.StartLine) - 1},
			                                ^
```

I would like to see all the typos to be fixed in 1 PR, if possible.

bvwells:
Hope #197 helps!

PR #197, opened 2019-10-03
Fix unconvert linting issues

bvwells:
Fix unconvert linting issues to address #187


bvwells:
Thanks for the fast merge @quasilyte!

PR #225, opened 2019-10-08
improve type inference

quasilyte:
An attempt to improve type inference.

```
Currently, if an expression has type T, it doesn't mean:
    "expression always have type T"
it only states that
    "expressions might have type T in some cases"
```

So, the function below is believed to return "int" type:

	function f($cond) {
		if ($cond) { return 10; }
		return $cond;
	}

The bad part is that it can return a value of type($cond).
The right answer would be `int|type($cond)`, but `int|mixed`
would still be better for now.

- try to fix misspelled types instead of just complaining about them

- make functions/expressions always have type.
  When precise type can't be inferred, use void/mixed.
  To avoid tons of new allocations, void and mixed type maps are shared
  as immutable type maps.

- move returnTypes from block context into the block itself,
  since now block checker lives for the entire function scope.
  It makes messing up with merging context impossible.
  We had several issues with type inference due to that
  (fixed in this change, tests are included).

- add left/right shifts type inference

### DIFF
--- a/src/solver/solver_test.go
+++ b/src/solver/solver_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func resolve(typ string) map[string]struct{} {
-	return ResolveType("", typ, make(map[string]struct{}))
+	return resolveType("", typ, make(map[string]struct{}))
 }
 
 func makeTyp(typ string) map[string]struct{} {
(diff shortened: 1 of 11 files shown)


quasilyte:
We can't yet say that `type=int` means that "type is always int", but this is a step forward in that direction.

One of the next steps would be separating different classes of type hints based on their guarantees.

src/solver/solver_test.go
```diff
@@ -50,10 +50,6 @@ func TestSolver(t *testing.T) {
 	meta.Info.AddFunctionsNonLocked("test", fm)
 	meta.Info.AddClassesNonLocked("test", cm)
 
-	if typ := resolve(meta.WrapFunctionCall(`\my_func`)); !typesEqual(typ, `array|bool|float`) {
```
YuriyNasretdinov: Why remove this?

quasilyte: Restored.

PR #226, opened 2019-10-08
add "make install"

quasilyte:
Make it easier for users to get a binary that include build info.
Also modify version printing a little bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ OS=`uname -n -m`
 AFTER_COMMIT=`git rev-parse HEAD`
 
 install:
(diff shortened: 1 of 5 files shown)


YuriyNasretdinov:
I would also change the name that is printed in version info to NoVerify (PHP Linter is the internal name in VK)

quasilyte:
Removed program name, so it's more universal now:
```
2019/10/08 15:32:25.877203 built on: 20191008153223 OS: lispbook x86_64 Commit: fe089f684fdbebfd5bc69a7c07b17dd07f70f60b
```
Also moved version printing to noverify main, so other linter builds can print version in their own way.

PR #237, opened 2019-10-10
resolve array to mixed[]

quasilyte:
So algorithms that infer array element by stripping [] suffix work
without ad-hoc elemOf("array")=>mixed case.


YuriyNasretdinov:
Please consider fixing the issues that I pointed out.

quasilyte:
@YuriyNasretdinov, fixed in #239, thanks!

PR #244, opened 2019-10-13
Whisk/fix/void result check/188

### LINKED ISSUE #188: Check void functions result usage
## Code Example

```php
/** @return void */
function f() {}

$_ = f(); // Assigned void result.
```

## Actual Behavior

No warnings.

## Expected Behavior

Warning about void result usage.

Note: this is a harder part of #86, so we can implement it as independent patches.

whisk:
Should (partially) fix #188 

Added a check when trying to assign void result to a single variable.

### DIFF
--- a/src/linttest/basic_test.go
+++ b/src/linttest/basic_test.go
@@ -204,7 +204,7 @@ func TestCallStaticParent(t *testing.T) {
 	runFilterMatch(test, "callStatic")
 }
 
-func TestVoidResultUsage(t *testing.T) {
+func TestVoidResultUsedInAssignment(t *testing.T) {
 	test := linttest.NewSuite(t)
 	test.AddFile(`<?php
 	/**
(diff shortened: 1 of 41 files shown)


whisk:
Changed type inference on assignments slightly to it supports method calls.
Also added check for binary ops.

src/linttest/basic_test.go
```diff
@@ -23,6 +23,72 @@ func TestCallStaticParent(t *testing.T) {
...
+	test.AddFile(`<?php
+	/**
+	* @return void
+	*/
+	function f() {}
+
+	f() & 1;
+	f() | 1;
+	f() ^ 1;
+	//f() && true; // skipped due to "undefined: Use true instead of true" report
```
quasilyte: Related to #250.
For now, tests should declare builtin constants explicitly, like:
```
function define($_, $_) {}
define('false', 1 == 0);
define('true', 1 == 1);
```

whisk: Should be fixed.

quasilyte:
:tada:

PR #248, opened 2019-10-14
fixed plus (+) operator type inference for arrays

### LINKED ISSUE #246: Fix `+` operator type inference
https://www.php.net/manual/en/language.operators.array.php

In PHP, `+` also works for arrays.
It performs union operation and the result type should be inferred as `mixed[]`.
Now it inferred as `int` or `float`.

https://github.com/VKCOM/noverify/blob/fd68c0fe476187de6312548fb9155c69a03a89b9/src/solver/exprtype.go#L315-L316

`+` should be handled a little bit differently than other arithmetical operations.

whisk:
Should fix #246

### DIFF
--- a/src/meta/typesmap.go
+++ b/src/meta/typesmap.go
@@ -71,7 +71,7 @@ func (m TypesMap) IsEmpty() bool {
 
 // Equals check if two typesmaps are the same
 func (m *TypesMap) Equals(m2 *TypesMap) bool {
-	if m2 == nil {
+	if m.m == nil || m2 == nil || m2.m == nil {
 		return false
 	}
 	if len(m.m) != len(m2.m) {
--- a/src/meta/typesmap.go
+++ b/src/meta/typesmap.go
@@ -71,7 +71,19 @@ func (m TypesMap) IsEmpty() bool {
 
 // Equals check if two typesmaps are the same
 func (m *TypesMap) Equals(m2 *TypesMap) bool {
-	return m.String() == m2.String()
+	if m2 == nil {
+		return false
+	}
+	if len(m.m) != len(m2.m) {
+		return false
+	}
+	for k := range m.m {
+		_, ok := m2.m[k]
+		if !ok {
+			return false
+		}
+	}
+	return true
 }
 
 // Len returns number of different types in map
--- a/src/meta/typesmap.go
+++ b/src/meta/typesmap.go
@@ -74,6 +74,30 @@ func (m TypesMap) Len() int {
 	return len(m.m)
 }
 
+// IsInt checks if map contains only int type
+func (m *TypesMap) IsInt() bool {
+	return m.Is("int")
+}
+
+// IsString checks if map contains only string type
+func (m *TypesMap) IsString() bool {
+	return m.Is("string")
+}
+
+// IsArray checks if map contains only array of any type
+func (m TypesMap) IsArray() bool {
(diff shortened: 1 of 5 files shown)


src/solver/exprtype.go
```diff
@@ -14,13 +14,23 @@ import (
...
 		return meta.NewTypesMap("int")
 	}
 	return meta.NewTypesMap("float")
 }
 
+// binaryPlusOpType is a special case as "plus" is also used for array union operation
+func binaryPlusOpType(sc *meta.Scope, cs *meta.ClassParseState, left, right node.Node, custom []CustomType) *meta.TypesMap {
+	// TODO: PHP will raise fatal error if one operand is array and other is not, so we may check it too
+	if ExprTypeLocalCustom(sc, cs, left, custom).IsArray() && ExprTypeLocalCustom(sc, cs, right, custom).IsArray() {
+		return meta.NewTypesMap("mixed[]")
```
YuriyNasretdinov: You might want to return a union of concrete types here instead so that information about types isn't lost

src/meta/typesmap.go
```diff
@@ -311,6 +311,11 @@ func (m *TypesMap) IsEmpty() bool {
 	return len(m.m) == 0
 }
 
+// Equals check if two typesmaps are the same
+func (m *TypesMap) Equals(m2 *TypesMap) bool {
```
YuriyNasretdinov: This is a very inefficient way to compare types.

src/solver/exprtype.go
YuriyNasretdinov: I would still suggest that you return the _union_ of the types (there is a MergeTypesMap method that does that) instead of using mixed.

Imagine the following (correct) PHP code:

<?php
class A { public function doSomething() { } }
class B { public function doSomething() { } }

$arrA = [ 'a' => new A(); ];
$artB = [ 'b' => new B(); ];
$arrUnion = $arrA + $arrB;

$arrA['a']->doSomething(); // no error
$arrB['b']->doSomething(); // no error
$arrUnion['a']->doSomething(); // this will be "no such method" mixed->doSomething () if you didn't merge the types

whisk: MergeTypesMap returns "int[]|string[]" for [1] + ["foo"], which is wrong. Correct type for this union should be (int|string)[].

Not sure if code supports "(A|B)[]" types yet.

YuriyNasretdinov: You're right but this is how noverify works right now. It currently does treat a[]|b[] as (a|b)[] anyway

src/meta/typesmap.go
whisk: Should be fixed.

YuriyNasretdinov: m can also be nil. There are no checks for this in String() method probably just because I didn't use it anywhere (it is there for debugging purposes but there is comment about that sadly).

YuriyNasretdinov: You can iterate nil maps and get their length (it will be 0). So you only need to check that m or m2 are nil. Two empty type sets can be equal as well so it's even incorrect to return false when they are nil. Also m.m will still cause panic if m is nil.

whisk: Updated. As of #257 we are not using pointers here which simplifies things a little.

quasilyte:
:tada:

PR #254, opened 2019-10-19
support for Alt* statement forms

### LINKED ISSUE #253: Handle isset/instanceof inside while cond as in if cond
## Code Example

```php
while (isset($x)) {
  // use of $x inside a loop
}
```

## Actual Behavior

Warning about undefined `$x`.

## Expected Behavior

No warnings.

instanceof also doesn't work inside `while (COND)` context.

quasilyte:
Handle Alt* forms like normal statement forms.
Maps AltFor->For, AltWhile->While, etc.

Fixes #253

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -358,6 +358,23 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
 		b.handleContinue(s)
 	case *binary.LogicalOr:
 		res = b.handleLogicalOr(s)
+
+	case *stmt.AltForeach:
+		s2 := stmt.Foreach(*s)
+		res = b.handleForeach(&s2)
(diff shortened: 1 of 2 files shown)


quasilyte:
> Relying on the fields of Alt* structures to be the same or even substituting them is very fragile and prone to errors.

If fields are not compatible Go wouldn't compile the code.

quasilyte:
> There can be cases where the handling would need to be different but the current solution doesn't allow that to happen without much pain.

Could you give an example?

quasilyte:
Will not be an issue after (and if) #255 is merged.

I merged alt* forms into normal forms while added `AltSyntax` flag to them, so they are still distinguishable, but by default, linter can handle them transparently.

PR #257, opened 2019-10-25
always use non-ptr version of meta.TypesMap

quasilyte:
TypesMap is 16 bytes on 64-bit systems (8 on 32-bit), there is
no point in passing it around as a pointer, since we already
have Append-like functions returning a new version of TypesMap objects.

The main benefit is that we don't need to guard against nil TypesMap anymore.

### DIFF
--- a/src/langsrv/defs.go
+++ b/src/langsrv/defs.go
@@ -26,7 +26,7 @@ type definitionWalker struct {
 	foundScopes []*meta.Scope
 }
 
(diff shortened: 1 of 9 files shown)


YuriyNasretdinov:
The only reason to do this was so that gob encoding worked. This may or may not break cache now.

quasilyte:
I'll check the cache again, but it seemed to work. (Not very trivial to verify that it's working correctly though.)

PR #259, opened 2019-10-25
meta: add TypesMap.Find method

quasilyte:
Some TypesMap.Iterate usages have a pattern of returning early
when they satisfy some predicate. This can be expressed more
elegantly with Find() method, where we have the next advantages:

- Caller can stop traversal with returning "true", no need to
  guard against "already found" in the beginning of the callback.
- No sorting is applied, since these tasks don't require
  predictable traversal to be deterministic.


YuriyNasretdinov:
It needs to be sorted. Otherwise you can find different methods each time if several options are possible. This leads to false diff messages that are especially bad when using full diff mode.

quasilyte:
Hm. I'll try to reproduce it and add tests that will fail (after running for high N times).

PR #266, opened 2019-10-28
linter: unexport some root/block methods

quasilyte:
These methods are only invoked through context that
doesn't need them to be exported.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -74,43 +74,6 @@ type BlockWalker struct {
 	nonLocalVars map[string]struct{} // static, global and other vars that have complex control flow
 }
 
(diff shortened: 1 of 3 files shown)


YuriyNasretdinov:
The methods were exported so that they can be used in custom linters. Did you replace it with something else? I don't see it here.

quasilyte:
They previously were needed because BlockContext/RootContext were interfaces => block/root walkers had exported methods that were called through the interface.

Now contexts are structs that contain pointers to walkers. Every method that was inside an interface is a method defined on these structs. Since we don't need to satisfy interfaces anymore, no exported methods are needed.

YuriyNasretdinov:
Ah, ok, thanks

PR #268, opened 2019-10-28
linter: fix class static prop fetch handling for $$

quasilyte:

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1022,11 +1022,8 @@ func (b *BlockWalker) handleStaticPropertyFetch(e *expr.StaticPropertyFetch) boo
 
 	sv, ok := e.Property.(*node.SimpleVar)
 	if !ok {
-		if vv, ok := e.Property.(*node.Var); ok {
-			vv.Expr.Walk(b)
-		} else {
-			e.Property.Walk(b)
-		}
+		vv := e.Property.(*node.Var)
+		vv.Expr.Walk(b)
 		return false
 	}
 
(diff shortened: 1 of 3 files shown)


YuriyNasretdinov:
I don't really see how it fixes the test. If I were to do the following, would it show any errors?

```
class Foo {}
function foo() {
	$x = "propname";
	return Foo::$$y; // $y is undefined, but $x is defined
}
```

quasilyte:
The only difference is that it does not walk the property even if it's not a var.
Old code looks like this:
```go
	varExpr, ok := e.Property.(*expr.Variable)
	if !ok {
		return false
	}

	varName, ok := varExpr.VarName.(*node.Identifier)
	if !ok {
		varExpr.VarName.Walk(b)
		return false
	}
```
New code:
```go

	sv, ok := e.Property.(*node.SimpleVar)
	if !ok {
		if vv, ok := e.Property.(*node.Var); ok {
			vv.Expr.Walk(b)
		} else {
			e.Property.Walk(b)
		}
		return false
	}
```
Can be changed to this to avoid changes in the behavior:
```go
	sv, ok := e.Property.(*node.SimpleVar)
	if !ok {
		if vv, ok := e.Property.(*node.Var); ok {
			vv.Expr.Walk(b)
		}
		return false
	}
```

quasilyte:
php7 grammar suggests that it's always some kind of var. So doing unchecked type assert is safe there.

YuriyNasretdinov:
Although I still don't quite understand why you need this change at all then:

```
- e.Property.Walk(b)
+ vv := e.Property.(*node.Var)
+ vv.Expr.Walk(b)
```

Isn't just walking the Property sufficient then?

YuriyNasretdinov:
Oh, I guess it's handling the "prop::$$x" case then. Yeah, that's probably good. Thanks

Issue #272, opened 2019-10-31
Add anonymous classes support


quasilyte:
This is a tracking issue for anonymous classes support.

Reported issues:
* #62

PR #273, opened 2019-10-31
fix bugs found during php-corpus check

quasilyte:
Running NoVerify on the big open source code corpus
revealed several bugs:

1. GetRootNode now returns *node.Root instead of node.Node,
   we assigned that to node.Node vairable and nil check
   was broken, causing linter to panic on nil nodes
   instead of giving "empty root node" notice.

2. Calling EnterNode on anon class gives a panic,
   since class name is nil and we can't set
   current class name because of that.
   Since a lot of class handling relies on
   current class name being available,
   skip anon class handling for now by
   returning false for these nodes.
   This cludge can be removed when #272 is resolved.

3. We can't eagerly resolve ClassConstFetch if
   ExprType is called when indexing is incomplete
   as it uses FindClass function that depends on meta.Index
   being complete. This causes race condition in fact
   (concurrent map reads and writes). Fixed by
   introducing wrapped class const fetch type.

### DIFF
--- a/src/linter/cache.go
+++ b/src/linter/cache.go
@@ -22,7 +22,8 @@ import (
 //     27 - added Static field to meta.FuncInfo
 //     28 - array type parsed as mixed[]
 //     29 - updated type inference for ClassConstFetch
-const cacheVersion = 29
+//     30 - resolve ClassConstFetch to a wrapped type string
+const cacheVersion = 30
 
 var (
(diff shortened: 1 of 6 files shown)


quasilyte:
Merging now, since it's a panic/race condition bug fix.
I can address possible review comments afterwards.

PR #275, opened 2019-10-31
rules: make it possible to have several patterns for 1 phpdoc comment

quasilyte:
When several patterns have exactly the same phpdoc comment, we want
to refactor that or to re-use constraints in some way.

Proposed way for that is phpdoc combining.

If a pattern doesn't have its own phpdoc comment, it borrows
all attributes from the last pattern with explicit phpdoc comment.

A good usage example of this feature is demonstrated in tests.

Made named rules enabled ("allowed") by default as well.
Whey can be disabled with -exclude-checks option if desired.

### DIFF
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -332,7 +332,7 @@ func initRules() error {
 		appendRules(linter.Rules.Root, rset.Root)
 		appendRules(linter.Rules.Local, rset.Local)
 
-		for _, name := range rset.AlwaysAllowed {
+		for _, name := range rset.ToAllow {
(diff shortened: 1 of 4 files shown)


quasilyte:
Hmm.. Makes sense. Thanks! I'll change it to your proposed syntax.

quasilyte:
We might as well consider whether it's OK that it makes it impossible to have a block as a pattern.

This is a valid pattern for finding all blocks that end with exit call:

```
{ ${"*"}; exit($_); }
```

quasilyte:
Will be done from scratch later. We have the #276 issue that'll preserve this PR as a reference.

PR #279, opened 2019-11-01
docs: restructured README

sergey-shambir:
User Guide now contains separate markdown pages to simplify navigation


sergey-shambir:
You can preview new README here: https://github.com/sergey-shambir/noverify/tree/restructured-readme
I'm sure that "Default lints" and "Custom lints" must be moved to separate page and mentioned under "User Guide" section, but we can do it later.

PR #294, opened 2019-11-02
linter: handle /* @var */ in addition to /** @var */

### LINKED ISSUE #284: undefined variable for @var-defined var
## Code Example

From yii2:

```php
<?php
/**
 * Creates a call for the method `yii\db\Migration::createTable()`
 */
/* @var $table string the name table */
/* @var $tableComment string the comment table */
?>        $this->addCommentOnTable('<?= $table ?>', '<?= $tableComment ?>');
```

## Actual Behavior

Undefined variable warnings for both variables.

```
<critical> ERROR   undefined: Undefined variable: table at /home/isharipov/CODE/php/basic/vendor/yiisoft/yii2/views/_addComments.php:7
?>        $this->addCommentOnTable('<?= $table ?>', '<?= $tableComment ?>');
                                        ^^^^^^
<critical> ERROR   undefined: Undefined variable: tableComment at /home/isharipov/CODE/php/basic/vendor/yiisoft/yii2/views/_addComments.php:7
?>        $this->addCommentOnTable('<?= $table ?>', '<?= $tableComment ?>');
                                                         ^^^^^^^^^^^^^
```

## Expected Behavior

Since PhpStorm permits this, maybe NoVerify can as well.
It does "jump" to `@var` annotations when trying to find a variable declaration.

### LINKED ISSUE #289: Support of /* @var ... */
Right now NoVerify only handle `/** @var T $var */`, but it seems like `/* @var T $var */` is also widely used and understood by PhpStorm.

If we don't support it, NoVerify loses types info and reports about undefined things that are in fact defined (if phpdoc info is correct).

Some numbers:

```
isharipov@lispbook: yiisoft$ grep -nr '/\*\* @var' | wc -l
42
isharipov@lispbook: yiisoft$ grep -nr '/\* @var' | wc -l
391
```

image

quasilyte:
Even though phpdoc documentation does not explicitly permit
/* @var */ to be used instead of /** @var */, there are
occasions where former form is used. PhpStorm does understand
it, so we can do the same at least by default.

It solves false positives that appear when we miss @var
info due to malformed syntax.

Fixes #289
Fixes #284

### DIFF
--- a/src/linttest/regression_test.go
+++ b/src/linttest/regression_test.go
@@ -6,6 +6,18 @@ import (
 	"github.com/VKCOM/noverify/src/linttest"
 )
 
+func TestIssue289(t *testing.T) {
+	linttest.SimpleNegativeTest(t, `<?php
+class Foo { public $value = 11; }
+
+$xs = [0, new Foo()];
+
+/* @var Foo $foo */
+$foo = $xs[1];
+$_ = $foo->value;
+`)
+}
+
 func TestIssue1(t *testing.T) {
 	linttest.SimpleNegativeTest(t, `<?php
 	interface TestInterface
--- a/src/phpdoc/parser.go
+++ b/src/phpdoc/parser.go
@@ -1,6 +1,8 @@
 package phpdoc
 
-import "strings"
+import (
+	"strings"
+)
 
 type CommentPart struct {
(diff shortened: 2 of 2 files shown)


quasilyte:
Accidentally mentioned someone with `@var` annotation. :sweat:

PR #304, opened 2019-11-09
Various small fixes

quasilyte:
Commits:
```
linter: simplify parsePHPDocVar method
linttest: remove BadModifiers test
phpgrep: fix input param accidental shadowing
meta: remove commented-out debug print stmt
all: fix style issues found by linters
```

### DIFF
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -627,11 +627,7 @@ func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
 
 		nm := p.Variable.Name
 
-		typ, errText := d.parsePHPDocVar(p.PhpDocComment)
-		if errText != "" {
-			d.Report(p.Variable, LevelWarning, "phpdocLint", errText)
-		}
-
+		typ := d.parsePHPDocVar(p.PhpDocComment)
 		if p.Expr != nil {
 			typ = typ.Append(solver.ExprTypeLocal(d.scope(), d.st, p.Expr))
 		}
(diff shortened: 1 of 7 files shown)


src/linter/root.go
```diff
@@ -627,11 +627,7 @@ func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
 
 		nm := p.Variable.Name
 
-		typ, errText := d.parsePHPDocVar(p.PhpDocComment)
-		if errText != "" {
-			d.Report(p.Variable, LevelWarning, "phpdocLint", errText)
-		}
-
+		typ := d.parsePHPDocVar(p.PhpDocComment)
```
YuriyNasretdinov: Why remove this check?

src/linttest/oop_test.go
```diff
@@ -104,23 +104,6 @@ $_ = WithProps::$int;
...
 }
 
-func TestBadModifiers(t *testing.T) {
-	t.Skip("Should be handled by other check, like keywordCase from #138")
-
-	test := linttest.NewSuite(t)
-	test.AddFile(`<?php
-class Foo {
-  /***/
-  PUBLIC Static function f() {}
```
YuriyNasretdinov: Why remove this test?

quasilyte: It's was always skipped and issues are covered in keywordCase test.

src/linter/root.go
quasilyte: It wasn't checking anything, parsePHPDocVar always returns an empty string as an error text.

PR #309, opened 2019-11-13
(experiment) Split linter block core checks

YuriyNasretdinov:
(experiment) Split linter block core checks to allow to add custom ch…ecks that only walk specific nodes in the tree.

This is analogous to the architecture suggested in ESLint, mentioned in the article by Instagram: https://instagram-engineering.com/static-analysis-at-scale-an-instagram-story-8f498ab71a0c

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -2,6 +2,7 @@ package linter
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


YuriyNasretdinov:
There is a ~4-5% performance penalty for doing this for linter core:

```
# full check time for a real large project

old:
20.50 real       142.88 user         9.70 sys

new:
21.57 real       149.69 user         9.62 sys

diff:
(149.69+9.62) / (142.88 + 9.70) = 1.044 (+4.4%)
```

quasilyte:
It's possible to do that without reflect if you use method values. But it's also OK to start with this if it makes it easier to decouple block walker even more.

For example:
```go
b := &BlockWalker{}
handleFor := b.handleFor
// handleFor has func(*stmt.For) type
```

I think it would be perfect to eventually implement all checks using "custom" API and make block/root walker only handle state like scopes and metadata.

YuriyNasretdinov:
> It's possible to do that without reflect if you use method values. But it's also OK to start with this if it makes it easier to decouple block walker even more.
Yeah but in order to support plugins callbacks we need to check all possible node.Node types and have a gigantic switch/case or something like that. Even though it might be faster, it leads to a lot of code. My approach requires to list all the methods explicitly for a plugin (and for core), which is still less work than trying to list all types IMHO.

quasilyte:
I think we should concentrate on the convenience of adding new checks rather than convenience of re-writing a framework. But since it's impossible to decouple everything right away and it will take more efforts, we can go with your proposed idea if you find it sustainable.

PR #330, opened 2020-01-06
linter: add misspell checker

### LINKED ISSUE #67: Add basic misspellings checks
## Code Example

```php
$enalbe = true;
```

## Actual Behavior

No warnings.

## Expected Behavior

Warning about a typo in `enalbe` (should be `enable`).

See https://github.com/client9/misspell.

We can start from checking only function parameter names.

quasilyte:
Uses github.com/client9/misspell for misspellings detection.
It's far from being comprehensive, but it finds some common
misspellings without false positives.

That is the main difference with PhpStorm spellchecking:
we give far less false positives and don't try to analyze
abbreviations or do any kind of real NLP.

This check runs spellchecker inside limited context only:

* Class/trait/interface names
* Method and function names
* Parameter names
* Class properties
* Constant names
* PhpDoc texts

It's also capable of detecting misspellings in _-separated identifiers.

misspell library is smart enough to avoid false positives
in hostnames, email addresses, and some other cases.

Fixes #67

Signed-off-by: Iskander (Alex) Sharipov <quasilyte@gmail.com>


quasilyte:
TODO:
* check on a bigger PHP projects to verify that we don't have major false positive cases
* check that `misspell.Replacer` is thread safe (otherwise we need to create a replacer for every root checker)

quasilyte:
Updates:
* split `misspell` into `misspellName` and `misspellComment` to make it easier to disable either of them if needed. It also adds some context to the warning message (misspell location).
* instead of `eng-local` use `misspell-list` parameter, so we can support custom sets if needed. `Eng` set is a generic misspelling set, `Eng,Eng/US` would also include US-specific dictionary.
* Added heuristics that avoid problems with interface names like `Responsable` which does look like a misspelling of `Responsible`, but it's a result of "-able" suffix convention for interfaces.

Checked it on several big PHP projects and now all results look like real misspellings without false positives.

PR #334, opened 2020-01-16
Adding an option to print information about files that take a long ti…

YuriyNasretdinov:
…me to parse and that can potentially prevent from being able to paralellize efficiently


src/cmd/flags.go
```diff
@@ -123,6 +125,7 @@ func bindFlags() {
 
 	flag.BoolVar(&linter.CheckAutoGenerated, `check-auto-generated`, false, "whether to lint auto-generated PHP file")
 	flag.BoolVar(&linter.Debug, "debug", false, "Enable debug output")
+	flag.StringVar(&debugParseDuration, "debug-parse-duration", "0ms", "Print files that took longer than the specified time to analyse")
```
quasilyte: It looks like it can be `flag.DurationVar(&linter.DebugParseDuration, )`, unless I'm missing something.

YuriyNasretdinov: It can if it exists... For some reason I didn't think it does.

YuriyNasretdinov: Fixed

Issue #347, opened 2020-01-27
Flag to disable exclusion of changes to master


setpill:
## Code Example

We use `noverify` with the following command in our CI setup:

```
- noverify \
	-git=.git \
	-git-skip-fetch \
	-git-commit-from=`git merge-base origin/master $COMMIT_SHA` \
	-git-commit-to=$COMMIT_SHA \
	-git-ref=refs/heads/$BRANCH_NAME \
	-cache-dir=$HOME/tmp/cache/noverify
```

## Actual Behavior

`Could not compute merge base between ORIGIN_MASTER and $LATEST_MASTER_COMMIT`

## Expected Behavior

For it to work without complaining. Seems it is trying to be too smart by (according to [the docs](https://github.com/VKCOM/noverify/tree/2f81ae701ae5b3b932d1d2480b3838f1223a5a91https://github.com/VKCOM/noverify/blob/master/docs/linter-usage.md)) "excluding changes made to master branch that is fetched to ORIGIN_MASTER". We want to use this tool to check files changed in a MR - regardless of what happened on master since. Therefore, files changed on master do not need to be excluded.

YuriyNasretdinov:
You can already achieve that by specifying `-git-ref=refs/heads/master` instead of `-git-ref=refs/heads/$ref` but I agree that providing an explicit option would be better.

Initially we didn't have exclusion of changes made in master but it made it impossible for us to enable noverify gradually for certain teams / people because if in the MR/PR there were merges from master that caused warnings they would also be shown during the push. As noverify at VK is set up as a blocking pre-push hook it was a problem so excluding master by default was the solution that was chosen.

setpill:
I tried your suggestion but it gives the same error. Included below is the command as run:

```
$ noverify -git=.git -git-skip-fetch -git-commit-from=`git merge-base origin/master $COMMIT_SHA` -git-commit-to=$COMMIT_SHA -git-ref=refs/heads/master -cache-dir=$HOME/tmp/cache/noverify
```

YuriyNasretdinov:
Hm, it turns out that this logic was removed by myself some time ago so you currently can't disable it. So I am not sure, but probably the best that you can do now is to make ORIGIN_MASTER to point to your `-git-from-commit` value so that no changes are excluded. From my understanding it should be something like ```
git update-ref ORIGIN_MASTER `git merge-base origin/master $COMMIT_SHA`
```

[...]

PR #348, opened 2020-01-28
Add flag to disable exclusion of changes on master

setpill:
Followup on #347

### DIFF
--- a/src/cmd/flags.go
+++ b/src/cmd/flags.go
@@ -21,15 +21,16 @@ var (
 
 	pprofHost string
 
-	gitCommitFrom       string
-	gitCommitTo         string
-	gitRef              string
-	gitPushArg          string
-	gitAuthorsWhitelist string
-	gitWorkTree         string
-	gitSkipFetch        bool
-	gitFullDiff         bool
-	gitIncludeUntracked bool
+	gitCommitFrom              string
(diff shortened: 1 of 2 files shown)


quasilyte:
Looks good to me.
@YuriyNasretdinov, what do you think?

YuriyNasretdinov:
I expected that -git-ignore-master flag would ignore changes from master, not ignore "ignoring changes from master". I would rename the flag to `-git-exclude-origin-master-changes` and set it to true by default. Yes, it is a long flag name but at least it's not misleading.

Otherwise the implementation is probably ok but of course we probably want to have some tests for this if we are ever going to change this logic.

setpill:
Yes I struggled a bit with the naming. Will rethink.

setpill:
Flag renamed to `-git-disable-compensate-master`. Hopefully that, combined with the help text, is clear in meaning.

quasilyte:
Thank you for the contribution!

YuriyNasretdinov:
We would probably also want to document the `-git-*` flags better as well actually :).

YuriyNasretdinov:
But that is a separate issue from this PR.

YuriyNasretdinov:
Created https://github.com/VKCOM/noverify/issues/349 to track that

PR #359, opened 2020-02-26
solver: add __call return type inference

### LINKED ISSUE #358: Preserve __call() types info
## Code Example

```php
class Magic {
  public function __call() { return $this; }
  public function notMagic() { return 10; }
}

$m = new Magic();
$m->magic1()->magic2()->notMagic();
```

## Actual Behavior

`$m->magic1()` inferred to `mixed` and "undefined method {mixed}->magic2` is reported.

## Expected Behavior

No warnings.

quasilyte:
We already do something similar for `__get()` methods to resolve unknown property fetches. This patch implements the same trick for `__call()`. It is only applied for the method calls that were not resolved by a direct method name (0 result types found for a given `methodName`).

Fixes #358


quasilyte:
TODO:
* Add __callStatic support?
* Add `@method` annotation support.

src/solver/solver.go
```diff
@@ -117,14 +128,12 @@ func (r *resolver) resolveTypeNoLateStaticBinding(class, typ string) map[string]
...
-		for className := range r.resolveType(class, expr) {
-			info, _, ok := FindMethod(className, methodName)
-			if ok {
-				for tt := range r.resolveTypes(className, info.Typ) {
-					res[tt] = struct{}{}
-				}
-			}
+		instanceTypes := r.resolveType(class, expr)
+		r.collectMethodCallTypes(res, instanceTypes, methodName)
+		if len(res) == 0 {
```
YuriyNasretdinov: I am fairly certain that `res` is going to be always empty here as there is no code that updates it anymore.

YuriyNasretdinov: Oh, I now see that `collectMethodCallTypes` actually fills in res. I would suggest to not modify the result map this way, or, for example, return it back (and call the method like "appendMethodCallTypes" instead) so that is more easily noticable.

src/linter/root.go
```diff
@@ -833,63 +838,115 @@ func (d *RootWalker) reportPhpdocErrors(n node.Node, errs phpdocErrors) {
 	}
 }
 
-func (d *RootWalker) parsePHPDocClass(doc string) classPhpDocParseResult {
-	var result classPhpDocParseResult
+func (d *RootWalker) parsePHPDocClassMethod(result *classPhpDocParseResult, part phpdoc.CommentPart) {
```
YuriyNasretdinov: I would write separate (unit?) tests for this piece of functionality as it seems complex enough and it is relatively self-contained.

src/solver/solver.go
quasilyte: I think my main reason to do that was a fact that we already allocate a map on a caller side, so it felt wasteful to allocate a new one and to re-assign that to a caller-side map.

I agree that this is not worthwhile. :)
I'll fix that to make code more easy to read.

YuriyNasretdinov:
I don't think we need to allocate a new map in this case, I would just suggest to make it more obvious that we are modifying it, we can return the same map back every time.

quasilyte:
Tests revealed an unpleasant bug.

If we suggest a `boolean -> bool` replace, we give up on phpdoc parsing and get nothing, but the entire point of `phpdocTypeFixer` is that it still returns some valid type along with that warning.

src/linter/root.go
quasilyte: Done.

Issue #363, opened 2020-03-20
if-related invariants not erased before handling else branch


quasilyte:
## Code Example

```php
<?php

class C {
  /***/
  public function example() {}
}

function f($c) {
  if ($c instanceof C) {
  } else {
    $c->example();
  }
}

```

## Actual Behavior

No warnings.

## Expected Behavior

Calling to undefined method, since `$c` is **not** C-typed.

quasilyte:
This is not only about instanceof but rather about all things recorder from `if`.

```php
function f() {
  if (isset($x)) {
  } else {
    var_dump($x);
  }
}
```
This code gives no warning about undefined `$x`.

quasilyte:
Another observation:

```php
if (isset($x) && $foo = 10) {}

var_dump($foo);
```

The code above does not generate any warnings because we treat everything in `if` condition as evaluated unconditionally.

It should probably report usage `$foo` as "might not be defined".

i582:
Done by #1058

PR #364, opened 2020-03-20
linter: improve method_exists handling

### LINKED ISSUE #362: Improve detection with class_exists/method_exists cases
I use that kind of checks (with `class_exists`+`method_exists` combo) regularly in my libraries to decide what algorithm to use. And i don't want to use `-exclude-checks` because it skips really bad cases in project.

## Installation and steps to reproduce

```bash
# I'm using osx
brew install go
# Installing noverify
mkdir -p $GOPATH/github.com/VKCOM
git clone https://github.com/VKCOM/noverify.git $GOPATH/github.com/VKCOM
cd $GOPATH/src/github.com/VKCOM/noverify
make install
# Use it as a static linter
cd /path/to/symfony3/project
noverify -exclude='vendor/|tests/|tmp/|var/' .
```

## Code Example

For example, this is what i've found after checking in my Symfony v3 project in file `bootstrap.php`:

```php
$path = dirname(__DIR__).'/.env';

if (!class_exists(Dotenv::class)) {
    throw new RuntimeException('Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.');
}

$dotenv = new Dotenv();
if (method_exists($dotenv, 'loadEnv')) {
    $dotenv->loadEnv($path); // <–– Critical is here
}
```

## Actual Behavior

```
<critical> ERROR   undefined: Call to undefined method {\Symfony\Component\Dotenv\Dotenv}->loadEnv() at /path/to/symfony3/project/app/bootstrap.php:9
        $dotenv->loadEnv($path);
                 ^^^^^^^
```

## Expected Behavior

No error here must be :-)

quasilyte:
We collect a slice of `<object, methodName>` pairs
inside a block context.

New pair entries are added via `method_exists(object, methodName)`
upon the if statement entry. These entries are removed
upon that if statement exit.

This new addition does not affect the type inference.
We only permit calling method_exists-defined methods,
but their types will be `(...mixed)->mixed`.

Fixes #362

Should be merged after #363 is resolved.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -950,7 +950,11 @@ func (b *BlockWalker) handleMethodCall(e *expr.MethodCall) bool {
 	e.Method.Walk(b)
 
 	if !foundMethod && !magic && !b.r.st.IsTrait && !b.isThisInsideClosure(e.Variable) {
-		b.r.Report(e.Method, LevelError, "undefined", "Call to undefined method {%s}->%s()", exprType, methodName)
+		// The method is undefined but we permit calling it if `method_exists`
+		// was called prior to that call.
+		if !b.ctx.customMethodExists(e.Variable, methodName) {
+			b.r.Report(e.Method, LevelError, "undefined", "Call to undefined method {%s}->%s()", exprType, methodName)
+		}
 	} else {
 		// Method is defined.
 
@@ -1471,6 +1475,18 @@ type andWalker struct {
 
 func (a *andWalker) EnterNode(w walker.Walkable) (res bool) {
 	switch n := w.(type) {
+	case *expr.FunctionCall:
+		args := n.ArgumentList.Arguments
+		nm, ok := n.Function.(*name.Name)
+		if ok && meta.NameEquals(nm, `method_exists`) && len(args) == 2 {
+			obj := args[0].(*node.Argument).Expr
+			methodName := args[1].(*node.Argument).Expr
+			lit, ok := methodName.(*scalar.String)
+			if ok {
+				a.b.ctx.addCustomMethod(obj, unquote(lit.Value))
+			}
+		}
+
 	case *binary.BooleanAnd:
 		return true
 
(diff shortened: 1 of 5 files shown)


src/linter/block.go
```diff
@@ -1471,6 +1475,18 @@ type andWalker struct {
 
 func (a *andWalker) EnterNode(w walker.Walkable) (res bool) {
 	switch n := w.(type) {
+	case *expr.FunctionCall:
+		args := n.ArgumentList.Arguments
+		nm, ok := n.Function.(*name.Name)
+		if ok && meta.NameToString(nm) == `method_exists` && len(args) == 2 {
```
YuriyNasretdinov: I would suggest to use NameEquals to avoid unnecessary allocations in hot paths

quasilyte: Maybe we can re-write NameEquals to be more efficient?

I think it also does allocations (it does strings.Split which allocates a slice).

quasilyte: Right now it doesn't seem to be faster:
```
BenchmarkNameEquals-8  13470901 79.3 ns/op 16 B/op  1 allocs/op
BenchmarkToString-8    19985751 59.1 ns/op 16 B/op  1 allocs/op
```
```go
var theName = &name.Name{
	Parts: []node.Node{
		&name.NamePart{Value: `method_exists`},
	},
}

func BenchmarkNameEquals(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = NameEquals(theName, `method_exists`)
	}
}

func BenchmarkToString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = NameToString(theName) == `method_exists`
	}
}
```

YuriyNasretdinov: You're probably right, we need to optimise NameEquals if it does allocate memory. The reason why I still think that we should use NameEquals is that once we do optimise it, all the code that uses it will become faster. If we always use NameToString we can only optimise the use-case where there is only part. If there are two parts, for example, memory allocation will still be required.

quasilyte:
It has the same problems as other if-related contexts.
I think it doesn't hurt much to merge this and then fix #363 with tests that cover all affected linter parts.

PR #366, opened 2020-03-20
meta: make NameEquals alloc-free

quasilyte:
Benchstat results:
```
name                 old time/op    new time/op    delta
NameEquals/1part-8     76.2ns ± 2%    15.3ns ± 2%   -79.94%  (p=0.008 n=5+5)
NameEquals/3parts-8     174ns ± 1%      30ns ± 1%   -82.74%  (p=0.008 n=5+5)

name                 old alloc/op   new alloc/op   delta
NameEquals/1part-8      16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
NameEquals/3parts-8     48.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)

name                 old allocs/op  new allocs/op  delta
NameEquals/1part-8       1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
NameEquals/3parts-8      1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
```

Refs #364

### DIFF
--- a/src/meta/metainfo.go
+++ b/src/meta/metainfo.go
@@ -492,16 +492,14 @@ func NameEquals(n *name.Name, s string) bool {
 		return false
 	}
 
-	sParts := strings.Split(s, `\`)
-
+	rest := s
 	for i, part := range n.Parts {
-		p, ok := part.(*name.NamePart)
-		if !ok {
+		part := part.(*name.NamePart)
+		if !strings.HasPrefix(rest, part.Value) {
 			return false
 		}
-
-		if p.Value != sParts[i] {
-			return false
+		if i != len(n.Parts)-1 {
+			rest = rest[len(part.Value)+len(`\`):]
 		}
 	}
 
--- a/src/meta/metainfo.go
+++ b/src/meta/metainfo.go
@@ -495,10 +495,14 @@ func NameEquals(n *name.Name, s string) bool {
 	rest := s
 	for i, part := range n.Parts {
 		part := part.(*name.NamePart)
-		if !strings.HasPrefix(rest, part.Value) {
-			return false
-		}
-		if i != len(n.Parts)-1 {
+		if i == len(n.Parts)-1 {
+			if part.Value != rest {
+				return false
+			}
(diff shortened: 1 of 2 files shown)


src/meta/metainfo.go
```diff
@@ -492,16 +492,14 @@ func NameEquals(n *name.Name, s string) bool {
...
 	}
 
-	sParts := strings.Split(s, `\`)
-
+	rest := s
 	for i, part := range n.Parts {
-		p, ok := part.(*name.NamePart)
-		if !ok {
+		part := part.(*name.NamePart)
+		if !strings.HasPrefix(rest, part.Value) {
```
YuriyNasretdinov: This code is slightly incorrect. I think we also need to check two things:

1. That there is a character `\` after `part.Value` ends if it's not the final part (essentially that the prefix is `part.Value + "\\"`).
2.  If it is the final part we need to check that there are no more characters in the string.

Otherwise the following would be equal:

`a\b` = `a\bb` because we only check the prefixes
`a b\` = `a\b` because we don't check where is `\` symbol

quasilyte: Good catch, thanks!

I believe if we handle the last part separately with "==" instead of "hasPrefix", all issues go away.

I've added test cases that check whether the mentioned scenarios lead to expected results. 

> 1 `a\b` = `a\bb` because we only check the prefixes

Now we check the last part with `==`, should be fixed

> 2. `a b\` = `a\b` because we don't check where is \ symbol

This will fail in old code as well, since the second part will be checked as
`HasPrefix("b", "")`, which is false. In updated code it shouldn't be an issue as well, as `"b"!=""`.

> 1. That there is a character \ after part.Value ends if it's not the final part (essentially that the prefix is part.Value + "\\").

I can't make up an example that will pass NameEquals with something except `\` as a separator. It will cause either different numbers of parts (we have an early check for it) or difference during the next part comparison. If you have an example, I can add it to tests and then it will be easier to fix it.

Issue #337, opened 2020-01-20
Undefined variable (if...elseif statement)


panychek:
## Code Example

```php
<?php

$expr = rand(0, 1);
if ($expr) {
    $var = 'foo';

} elseif ($var = 'bar') {

}

echo $var;

```

## Actual Behavior

```
<critical> INFO    undefined: Variable might have not been defined: var at /var/www/dev1/www/test.php:11
echo $var;
     ^^^^
```

## Expected Behavior

No issues

PR #371, opened 2020-03-20
linter: improve if-related implicit vars handling

### LINKED ISSUE #363: if-related invariants not erased before handling else branch

### LINKED ISSUE #369: Using undefined variable causes it to be "conditionally defined"
## Code Example

```php
<?php

function f($cond) {
    if ($cond) {
        echo $x; // 1
    }
    echo $x; // 2
}

```

## Actual Behavior

```
<critical> ERROR   undefined: Undefined variable: x at hello.php:5
        echo $x; // 1
             ^^
<critical> INFO    undefined: Variable might have not been defined: x at hello.php:7
    echo $x; // 2
         ^^
```

## Expected Behavior

Variable should be reported as undefined twice.

### LINKED ISSUE #370: No warnings for if-else with instanceof chains
## Code Example

```php
<?php

class Fooo1 { public $a; }
class Fooo2 { public $b; }

global $globf;

if ($globf instanceof Fooo1) {
    echo $globf->a;
} elseif ($globf instanceof Fooo2) {
    echo $globf->a; // <-- no warnings (bad)
    echo $globf->b;
} else {
    echo $globf->a; // <-- no warnings (bad)
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Warning for undefined property `a`.

The problem is, noverify assigns `\Fooo1|\Fooo2` type for `$globf` inside a second if body, so we can access both Fooo1 and Fooo2 members there.

quasilyte:
This change attempts to improve handleIf behavior.

We introduce implicit kind of vars to express automatically
defined vars that can be handled in a special way.

andWalker.varsToDelete are removed because we're running
all branches in a separate block contexts. No extra
cleanup is required.

Changed code requires to evaluate if statement condition
separately, so we can get variables defined inside it
into the parent block context.
assignWalker type does that.

Fixes #369
Fixes #363
Fixes #370

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1462,15 +1462,10 @@ func (b *BlockWalker) propagateFlagsFromBranches(contexts []*blockContext, links
 	}
 }
 
-// andWalker walks if conditions and adds isset/!empty/instanceof variables
+// andWalker walks if conditions and adds isset/!empty/instanceof implicit variables
 // to the associated block walker.
-//
-// All variables defined by andWalker should be removed after
-// if body is handled, this is why we collect varsToDelete.
 type andWalker struct {
 	b *BlockWalker
-
-	varsToDelete []node.Node
 }
 
 func (a *andWalker) EnterNode(w walker.Walkable) (res bool) {
@@ -1498,16 +1493,14 @@ func (a *andWalker) EnterNode(w walker.Walkable) (res bool) {
 
 			switch v := v.(type) {
 			case *node.SimpleVar:
-				a.b.addVar(v, meta.NewTypesMap("isset_$"+v.Name), "isset", true)
-				a.varsToDelete = append(a.varsToDelete, v)
+				a.b.ctx.sc.AddImplicitVar(v, meta.NewTypesMap("isset_$"+v.Name), "isset", true)
 			case *node.Var:
 				a.b.handleVariable(v.Expr)
 				vv, ok := v.Expr.(*node.SimpleVar)
 				if !ok {
 					continue
 				}
-				a.b.addVar(v, meta.NewTypesMap("isset_$$"+vv.Name), "isset", true)
-				a.varsToDelete = append(a.varsToDelete, v)
+				a.b.ctx.sc.AddImplicitVar(v, meta.NewTypesMap("isset_$$"+vv.Name), "isset", true)
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


src/linter/block.go
```diff
@@ -1623,17 +1602,22 @@ func (b *BlockWalker) handleIf(s *stmt.If) bool {
 	linksCount := 0
 
 	if s.Stmt != nil {
-		linksCount += walk(s.Stmt)
+		// We evaluate s.Cond inside a new context to avoid
```
YuriyNasretdinov: There are some (rare) cases when it might actually matter. For example:

```
if ($a = do_something()) {
// ...
} else {
// ...
}
```

In this example `$a` will be always defined (in all contexts, including the else block) because it was assigned. It does not mean that all expressions in the first if statement will be executed though but the basic idea why I thought we need to execute the first if condition in the base context was in order to handle this scenario.

YuriyNasretdinov: I like your approach better anyway. It does indeed solve a lot of issues that you mentioned.

src/linter/block.go
```diff
@@ -1660,6 +1644,10 @@ func (b *BlockWalker) handleIf(s *stmt.If) bool {
 		b.ctx.sc.AddVarName(nm, types, "all branches", defCounts[nm] == linksCount)
 	}
 
+	// Collect all variable definitions from the if condition into the current context.
```
YuriyNasretdinov: Actually I don't think that we need to put _all_ assignments from the `if` condition into the current context. Also this code does not take into account that in a case like this:

```if($a = something()) { ... } else { ... }```

the `else` block would also have `$a` defined.

We only need to evaluate the first assignment in certain cases, for example:

```
if ( ($a = abc()) && ($b = bcd()) { ... } // $a is always defined, $b isn't
if ( ($a = abc()) || ($b = bcd()) { ... } // $a is always defined, $b isn't
if ($a = abc()) { ... } // $a is always defined
if (!$a = abc()) { ... } // // $a is always defined
```

And not in the other ones, for example in the following cases we don't want to handle any assignments:

```
if (abc() && ($b = bcd()) { ... } // $b might be defined
```

quasilyte:
In addition to all mentioned above, it would be nice to check if we can solve #337 as well.

quasilyte:
There is also a somewhat related issue #374.

i582:
There have been many changes, including in IR, so it makes no sense to keep PR open. All the same, there is not much sense in this, since it has been hanging unchanged for more than a year.

PR #385, opened 2020-04-11
makefile: add -race flag to "go test" command

quasilyte:
This way we have chances to catch some race conditions earlier.


quasilyte:
For some reason, Travis doesn't want to show the failure logs.

PR #386, opened 2020-04-11
linter: improve undef method handling of abstract classes

### LINKED ISSUE #315: Abstract classes that extend interface cause "undefined" errors
## Code Example

```php
<?php

interface LoggerInterface {
    public function Log($msg);
}

abstract class AbstractLogger implements LoggerInterface {
    public function LogError($msg) {
        $this->log('error: ' . $msg);
    }
}

class EchoLogger extends AbstractLogger {
    public function Log($msg) {
        echo "$msg\n";
    }
}

$l = new EchoLogger();
$l->LogError('123');
```

## Actual Behavior

```
<critical> ERROR   undefined: Call to undefined method {\AbstractLogger}->log() at example.php:9
        $this->log('error: ' . $msg);
               ^^^
```

## Expected Behavior

No warning.

quasilyte:
Permit calling unimplemented iface methods from the abstract classes.

Fixes #315


src/linter/block.go
```diff
@@ -960,6 +960,10 @@ func (b *BlockWalker) handleMethodCall(e *expr.MethodCall) bool {
 		magic = haveMagicMethod(typ, `__call`)
 		return foundMethod || magic
 	})
+	// We permit abstract classes to call unimplemented interface methods.
+	if !isAbstractClass(b.r.currentClassNode) && implClass == "" {
```
YuriyNasretdinov, line 964: I would change the order of the arguments, e.g. check for implClass == "" first because determining whether or not the class is abstract can be much slower. Also I think you need to only check calls to "$this->" and "self::" (maybe "parent::" too) and not everything because you can call different objects from inside an abstract class and it should still be an error if such methods don't exist. In general you need to check that the class of the _object_ that is being called is abstract and not the current class.

quasilyte: Good catch. I tried to fix that, PTAL.

PR #407, opened 2020-04-12
linter: add "unimplemented" checker

### LINKED ISSUE #368: Report undefined classes used in "extend" clause
## Code Example

```php
<?php

class Foo extends Undefined {}
```

## Actual Behavior

No warnings.

## Expected Behavior

Should warn about undefined class `Undefined`.
The same goes for undefined interfaces.

### LINKED ISSUE #395: Check that interfaces are implemented by (non-abstract) classes
## Code Example

```php
interface FooInterface {
  public function foo();
}

class Fooer implements FooInterface {
  private function bar() {
    return $this->foo();
  }
}
```

## Actual Behavior

No warnings.

(Well, undefined methods was reported before, but this might change with #386 so we need this diagnostic now more than before.)

## Expected Behavior

Report that `Fooer` does not implement `FooInterface`.

quasilyte:
For every interface in "implements" and class in "extends" run
a method set check.

That check is skipped for abstract classes.

Non-abstract classes must implement every abstract method of their
parents (directly or via traits) and every method from interfaces chain.

Related changes:

- Added FuncAbstract bit to FuncFlags

- solver.FindFunc now correctly prefers concrete method definitions
  instead of the closest one.

- Using undefined class in extend/implements clause cause "undefined" error.

Fixes #395
Fixes #368

### DIFF
--- a/src/langsrv/langsrv.go
+++ b/src/langsrv/langsrv.go
@@ -277,16 +277,16 @@ func handleTextDocumentSymbol(req *baseRequest) error {
 			filename := strings.TrimPrefix(uri, "file://")
 			res := meta.Info.GetMetaForFile(filename)
 
-			for className, classInfo := range res.Classes {
+			for _, classInfo := range res.Classes.H {
 				result = append(result, vscode.SymbolInformation{
-					Name:     baseSymbolName(className),
+					Name:     baseSymbolName(classInfo.Name),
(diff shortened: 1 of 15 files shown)


quasilyte:
I'll run it on some open-source projects and report back.

If everything goes well, we can try merging it.

PR #410, opened 2020-04-13
update phpstorm stubs

### LINKED ISSUE #409: Handle duplicates (overloadings) in phpstorm stubs
## Code Example

https://github.com/JetBrains/phpstorm-stubs/blob/b1041ca15222138a1d3d3aa4afb096df215bdc2e/standard/standard_1.php#L618

```php
function strtr ($str, $from, $to) {}
function strtr ($str, array $replace_pairs) {}
```

Here we have 2 definitions for `strtr`, but only first is recorded, so we report `strtr` usage with 2 arguments as erroneous (`argCount` asks for an extra argument).

I propose merging definitions of whether we can. In that case, we can infer args count to be in 2-3 range.

quasilyte:
Also added a heuristic that tries to choose min(MinParamsCnt)
among all function overloadings.

Such overloadings only occur in PhpStorm stubs:
	\strtr: 2 or 3 args
	\stream_context_set_option: 2 or 4 args

Fixes #409

### DIFF
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -1214,7 +1214,7 @@ func (d *RootWalker) enterFunction(fun *stmt.Function) bool {
 	if sideEffectFreeFunc(d.scope(), d.st, nil, fun.Stmts) {
 		funcFlags |= meta.FuncPure
 	}
-	d.meta.Functions[nm] = meta.FuncInfo{
+	meta.AddFunctionToMap(d.meta.Functions, nm, meta.FuncInfo{
 		Params:       params,
 		Pos:          d.getElementPos(fun),
 		Typ:          returnType.Immutable(),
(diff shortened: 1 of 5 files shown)


quasilyte:
Now we try to handle duplicates in both RootWalker and meta.Info.

src/linter/root.go
```diff
@@ -1214,15 +1214,15 @@ func (d *RootWalker) enterFunction(fun *stmt.Function) bool {
 	if sideEffectFreeFunc(d.scope(), d.st, nil, fun.Stmts) {
 		funcFlags |= meta.FuncPure
 	}
-	d.meta.Functions[nm] = meta.FuncInfo{
+	meta.AddFunctionToMap(d.meta.Functions, nm, meta.FuncInfo{
```
YuriyNasretdinov, line 1217: You're not holding the metadata lock here.

YuriyNasretdinov: Sorry, got confused. You're right, you're modifying the local map here. Maybe rename it to something like 'AppendFunction' and return the modified map so that it's clearer that you're not modifying the global state in meta?

quasilyte:
Since it turned out to be more complicated than I thought, I'll send the stubs update separately.

quasilyte:
stubs were updated in the separate PR.

PR #455, opened 2020-05-15
linter: add redundantGlobal checker

quasilyte:
`global $superglobal` is redundant and should not be used.

https://www.php.net/manual/en/language.variables.superglobals.php


quasilyte:
I extended the checker description.
The idea is that users can use `--help` to get information about every checker:
```
  redundantGlobal
    	Report global statement over superglobal variables (which is redundant).
```

PR #463, opened 2020-05-20
linter: add tuple type support

### LINKED ISSUE #436: Implement tuple() type support
Issue #424 describes fixed array types (also known as "shape" types).

A `tuple` is a shorthand for `shape` with numerical keys:

```
tuple(T) = shape{0:T}
tuple(T1,T2) = shape{0:T1,1:T2}
```

Note that `array{key:type}`  is identical to `shape{key:type}`.

Shape/fixed array types were implemented in a94488f08ae497bc9a57ee004517b694bd4bf7ed

Related code:
https://github.com/VKCOM/noverify/blob/056fedd989c5220c37470158609daa3698bafa79/src/linter/types.go#L66-L79

A new "tuple" case needs to be added.

Tests should be added to `src/linttest/shape_test.go` file.
In addition, more tests can be added to `src/linttest/exprtype_test.go` (see `TestExprTypeShape`).

ludweeg:
The `tuple` type is a shorthand for `shape` with numerical keys: 
`tuple(T1, T2) = shape{0:T1, 1:T2}`

Fix #436

### DIFF
--- a/src/linter/types.go
+++ b/src/linter/types.go
@@ -77,6 +77,10 @@ func (conv *phpdocTypeConverter) mapType(e phpdoc.TypeExpr) []meta.Type {
 		if typ.Value == "shape" {
 			return conv.mapShapeType(params)
 		}
+		if typ.Value == "tuple" {
+			return conv.mapTupleType(params)
+		}
+
 		return conv.mapType(typ)
 
 	case phpdoc.ExprNullable:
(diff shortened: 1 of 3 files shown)


src/linter/types.go
```diff
@@ -149,6 +153,34 @@ func (conv *phpdocTypeConverter) mapShapeType(params []phpdoc.TypeExpr) []meta.T
...
+
+		args := []phpdoc.TypeExpr{
+			phpdoc.TypeExpr{
+				Kind:  phpdoc.ExprInt,
+				Value: fmt.Sprint(i),
+			}, p}
+
+		typeExpr := phpdoc.TypeExpr{
+			Kind:  phpdoc.ExprKeyVal,
+			Value: fmt.Sprintf("%d:%s", i, p.Value),
```
quasilyte, line 175: I think `mapShapeType` doesn't need ExprKeyVal value. It only uses Key (args[0]) value to determine which array key shape type should match.

If it's not required, maybe we can avoid initializing it?

src/linter/types.go
```diff
@@ -149,6 +153,34 @@ func (conv *phpdocTypeConverter) mapShapeType(params []phpdoc.TypeExpr) []meta.T
...
+		if p.Kind != phpdoc.ExprName && p.Kind != phpdoc.ExprGeneric {
+			conv.warn(fmt.Sprintf("tuple param #%d: want type, found %s", i+1, p.Value))
+			continue
+		}
+
+		args := []phpdoc.TypeExpr{
+			phpdoc.TypeExpr{
+				Kind:  phpdoc.ExprInt,
+				Value: fmt.Sprint(i),
+			}, p}
```
quasilyte, line 171: How about this?
```go
key := phpdoc.TypeExpr{
	Kind:  phpdoc.ExprInt,
	Value: fmt.Sprint(i),
}
typ := p
args := []phpdoc.TypeExpr{key, typ}
```

src/linter/types.go
```diff
@@ -149,6 +153,34 @@ func (conv *phpdocTypeConverter) mapShapeType(params []phpdoc.TypeExpr) []meta.T
 	return []meta.Type{{Elem: shape.name}}
 }
 
+func (conv *phpdocTypeConverter) mapTupleType(params []phpdoc.TypeExpr) []meta.Type {
+	types := make([]phpdoc.TypeExpr, 0, len(params))
+	for i, p := range params {
+		if p.Value == "*" || p.Value == "..." {
+			continue
+		}
+		if p.Kind != phpdoc.ExprName && p.Kind != phpdoc.ExprGeneric {
```
quasilyte, line 162: Do we need this kind of restriction?

I think we can remove this check and leave all validation to `mapShapeType`.

src/linttest/shape_test.go
```diff
@@ -144,3 +144,43 @@ $s1 = new_shape1();
...
+
+$t = new T();
+echo $t->bad1['x']->value;
+
+echo $t->good2[0]->value;
+echo $t->good3[1][0]->value;
+`)
+	test.Expect = []string{
+		`tuple param #1: want type, found x:Box`,
+		`tuple param #2: want type, found x[]`,
```
quasilyte, line 181: See the other comment about type validation inside mapTupleType.
I think it's a good idea to do type node validation, but it should be done in mapShapeType so both tuple and shape types can be validated.

Also, `x[]` is a valid tuple/shape type element. :)

I would suggest adding validation code in a separate PR.

PR #468, opened 2020-05-21
linter: don't recurse when matching rules

quasilyte:
phpgrep Find() method is recursive.
It tries to match at the top leven, then it traverses
the AST and calls the callback for every matching node.

A new tests (added in this change) resulted in 2 warnings
instead of 1 due to this.

It's not enough to do "return false" to stop the traversal.

To avoid this confusion in future, Find is renamed to FindAll.
A new Find method does not perform any traversal, which is
exactly what we need.

More refactoring can be done later. This change goal
is to fix a bug.


quasilyte:
It never triggered before because it's very rare for the expression to have a matching expression within itself. It's much easier to see this behavior with statements.

PR #473, opened 2020-05-23
linter: add parentConstructor checker

### LINKED ISSUE #408: If base class has non-default constructor, check whether class constructor calls parent::__construct
## Code Example

```php
class A {
  public function __construct() {
    // Something clever.
  }
}

class B extends A {
  public function __construct($x) {
    $this->x;
    // No call to parent::__construct()
  }
  private $x;
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Warn that parent constructor was not called.

quasilyte:
parentConstructor reports __construct() methods that
do not call parent::__construct() when they need to do so.

We require a parent constructor call if a class extends
the other class that has a explicitly defined constructor
(which is callable from the derived class).

Fixes #408

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -71,6 +71,11 @@ type BlockWalker struct {
 	// When can't infer precise type, can use mixed.
 	returnsValue bool
 
+	// callsParentConstructor is set to true when parent::__construct() call
+	// is found. This is needed for a root walker to report constructors
+	// that do not call parent constructors.
+	callsParentConstructor bool
+
 	// shared state between all blocks
 	unusedVars   map[string][]node.Node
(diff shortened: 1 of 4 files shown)


quasilyte:
We check that the parent class have __construct method:

image

We don't record default constructor, so this check should be enough.

YuriyNasretdinov:
Sorry for being lazy to explain in full detail. I meant the following situation:

```
<?php
class A { public function __construct() { ... } }
class B extends A { } // should be no errors IMO
```

That is because oftentimes there is no need to redefine the constructor in child classes.

quasilyte:
I think we cover that case as well:
image
We only perform this check for the __construct method itself.

YuriyNasretdinov:
Ah, sorry, I misunderstood the tests then.

PR #476, opened 2020-05-25
linter: add 2 exceptions to type precision inference

quasilyte:
1. Parameter default value is not a precise source.

2. Static variables should be skipped as their assignment
   is only executed once (type can change after another assignment).


quasilyte:
This is not a complete solution.
Now I see that exprtype tests are not good enough as they only cover return type inference which can remove "precise" flag in many cases. This is not true for other cases where we don't want a type to stay precise. To fix that, we need tests that can actually reveal the issue and then ensure that the fix is not broken in the future.

PR #479, opened 2020-05-27
meta: add missing WElemOfKey case to formatType function

quasilyte:


quasilyte:
Not sure which string representation to use there tbh.

Issue #402, opened 2020-04-12
Add langserver tests


quasilyte:
It can be useful to make sure linter changes do not break the language server.

See #398

PR #487, opened 2020-05-29
langsrv: Improved usage of URI

edgardmessias:
Instead of only prepend "file://", this pull solve problem on windows platform, instead of "file://C:..." (double slash), it uses "file:///C:..." (triple slash)

Before: `file://c:\<user path>\.vscode\extensions\edgardmessias.php-noverify-0.1.0\stubs\Core\Core_c.php`
After: `file:///c:/<user path>/.vscode/extensions/edgardmessias.php-noverify-0.1.0/stubs/Core/Core_c.php`

Maybe, we can start use of:
* https://github.com/go-language-server/protocol
* https://github.com/golang/tools/tree/master/internal/lsp/protocol


quasilyte:
Could you do `git rebase master`, please?
I can press the "update branch" but it can mess up with your workflow if your prefer another way of updating your branch.

The submitted code looks good. I'll take another look when the branch is up to date.

edgardmessias:
@quasilyte done

YuriyNasretdinov:
I would also like to add that I would definitely be not against using some LSP server library instead of doing what we do now. I don't think it was an option when I started it sadly.

edgardmessias:
I never programmed in golang before this project, so I started learn two days ago.

I have a few knowledge about LSP. I worked a small project months ago using typescript.

src/langsrv/langsrv.go
```diff
@@ -94,6 +96,19 @@ func writeMessage(message interface{ IMessage() }) error {
 	return err
 }
 
+func isFileScheme(documentURI uri.URI) bool {
+	u, err := url.ParseRequestURI(string(documentURI))
```
quasilyte, line 100: It's unfortunate that `uri.URI` does not store scheme information after a `Parse`. :(

quasilyte:
If you're interested in a discussion of lang server testing, there is #402. :wink:

src/langsrv/langsrv.go
edgardmessias: Yes, it is a pain

PR #489, opened 2020-05-30
Migrated from dep to Go 1.12 modules (close #342)

### LINKED ISSUE #342: Migrate from dep to go mod
Hi,

This project is still using Gopkg.toml / Gopkg.lock from dep, please consider converting to the official solution Go Modules.

As project is still in v0 there should be no issues with the import-path,

edgardmessias:
Migrated from dep to Go 1.12 modules (close #342)


edgardmessias:
It is the same idea of composer or npm, instead of keep a copy of all files, you only need set the reference.
You can see the installed packages in `<go path>\pkg\mod`

mappu:
> I would still prefer to have all dependencies materialised in the repo though (the `vendor/` directory). I actually have never used Go modules myself but I think it should be allowed?

You can do it (run `go mod vendor` to rebuild the `vendor/` directory).

YuriyNasretdinov:
The reason to still have `vendor/` directory in your repo (I mean, committed in the repo) is to prevent authors of the original repos from deleting the branches/commits that you depend on and from breaking backwards compatibility. It also allows you to clone only a single repo instead of fetching hundreds or thousands of repos that you depend on. I have hit similar issues in the past (and not only myself, you can remember left-pad npm repo issue) and I don't think it's a high price to pay to always store `vendor/` inside to prevent issues that I mentioned from happening to (somewhat) popular repos.

Issue #325, opened 2019-12-12
Duplicated keys are not detected for non-literal keys


quasilyte:
## Code Example

```php
$foo = [C1 => 1, C2 => 2, C1 => 3];
```

## Actual Behavior

No warnings.

## Expected Behavior

Duplicate array key warning.

YuriyNasretdinov:
Only constant expressions should be counted though as there exist things like this (in theory):

```
<?php
[ rand() => 1, rand() => 2 ];
```

This code is not great but it usually does not contain duplicate keys :)

quasilyte:
We have `sideEffectFree(expr)` that can reject non-pure calls and other expressions that can have side effects. If it's not good enough, we can adjust it to be good enough.

quasilyte:
Maybe we can use nodeSet for this, but maybe it's still easier to support a few types of nodes:
* `expr.ConstFetch`
* `expr.ClassConstFetch`
* `scalar.Lnumber`
* `scalar.Dnumber`
* `scalar.String`
(Maybe something else?)

hazzus:
Guys, sorry about this spam, I thought that testing pull requests in my fork won't mention original repo issue. :smile: 
I will have my pull request now opened here

PR #495, opened 2020-06-03
Add reports for non-literal key duplicates in dupArrayKeys

hazzus:
Solving #325. Now separately looks at duplicates in `scalar.Dnumber`, `expr.ConstFetch`, `expr.ClassConstFetch`. Moreover, it adds to dupArrayKeys the check for any other `node.Node` satisfying `sideEffectFree` condition.
This PR also improves Lnumber duplicates check, considering they may be written in different formats. 
Example code:
```php
$example5 = [
  17 => 1,
  0x11 => 3, // duplicate key 17
  0b10001 => 2, // duplicate key 17
];
```
The only help needed: Does it make sense to use a `printer` in default (`sideEffectFree()`) case? I just didn't come up with any other idea (besides using a case switch for everything).

PS: This PR is a test task for the internship in VK in the backend-infrastructure team.


quasilyte:
Hello, Pavel!

Here is the agenda:

quasilyte:
> The only help needed: Does it make sense to use a printer in default (sideEffectFree()) case? I just didn't come up with any other idea (besides using a case switch for everything).

See the previous message. Please send me a mail with all your questions somewhere around the 15th of June and I'll answer them unless they're asking for a concrete code. :)

I could answer here, but we don't want to make it too easy for other applicants. :wink:

hazzus:
**Update**:
In last few commits I moved the implementation to NodeSet. It was considered as the solution in issue. Code is simpler, solution is more abstract, it avoids some overhead on checks.
However, this reverts checks of different formats for integers and casted doubles. As it was mentioned, this is a rare case.
I'm still not sure about `ValidArrayKey` check, it may be more completed in the future.

hazzus:
**Update**:
1. Moves tests from regression to basic
2. Removes redundant quotes in the check message
3. Extends `ValidArrayKey` check: now checks types of some nodes (oriented on `exprType` function). New version of validation doesn't check precision of typeMap, however, it checks that **all** possible types **must be** an arrays or objects in order not to create a false-positive output (so if `mixed`, `unknown_from_list` occurs, it satisfies them as valid keys)

quasilyte:

PR #503, opened 2020-06-04
linttest: shuffle PHP input files

quasilyte:
In our tests we don't do concurrent indexing/linting which
makes it harder to catch undesired processing order dependencies.

With randomized indexing/linting we're getting closer to the
real world cases where order of parsing is not guaranteed to
be the same between executions.

Test failure prints a seed that was used.
Using that value with `$TEST_SEED` environment variable
makes it possible to easily reproduce the failing state.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,6 @@ install:
 
 check:
 	@go vet $(go list ./src/... | grep -v vendor)
-	@go test -race -v ./src/...
+	@go test -count 3 -race -v ./src/...
 
 .PHONY: check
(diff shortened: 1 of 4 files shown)


quasilyte:
Seems like we need to sort results to make golden tests pass.

I also noticed that NoVerify does not sort reports, so its output is not stable between runs.

quasilyte:
Maybe integration tests that would run an actual NoVerify binary is not such a bad idea. :sweat_smile:

PR #505, opened 2020-06-06
example,wasm: make go build ./... succeed

quasilyte:
For wasm, we need all files to be excluded by the
build constraint (tags) to make go build/install/etc
pass on the repository root.

We also had leftovers of the main package in php/parser
which created a duplicated main package entry in the tree.

Another touch is fixing the example package.
"make check" will now run example/custom tests,
so hopefully we'll never break the example again.


quasilyte:
@YuriyNasretdinov do you remember what was the problem with having a `+build wasm` in the wasm folder? It looks like most commands work for me and we finally can run `./...` commands at the source root instead of doing `./src/...` which is less convenient. But I'm afraid that I'm forgetting something important.

YuriyNasretdinov:
The issue with +build wasm was that if you include it, you could not build the package by default as go would complain that there are no buildable files in the directory. But I assume that is not a problem anymore so it should be fine.

PR #506, opened 2020-06-08
Added non-literal key duplication report

DmiitriyJarosh:
This pull request solves #325. I have used sideEffectFree() function for detection of const keys and then format some of nodes to special format to handle PHP rules of array keys converting. As I think it is to complex for this task to implement the whole node tree switch, I have tried to simplify it maximally. As a result features of array last comma and different string quotes are implemented as tricks. I can't find the example which will destroy my trick. If you can, please, tell me about it. Also I have added test for this issue.

PS: This PR is a test task for the internship in VK in the backend-infrastructure team.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -3,6 +3,7 @@ package linter
 import (
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


quasilyte:
Hello @DmiitriyJarosh!

Here is the agenda:

Issue #324, opened 2019-12-12
Some class-scoped expressions are not analyzed


quasilyte:
## Code Example

```php
<?php
class C {
  public static $arr = [
    'key1' => 'something',
    'key2' => 'other_thing',
    'key1' => 'third_thing', // duplicate
  ];
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Warning about duplicated array key.

PR #507, opened 2020-06-09
Adding a check for some class fields

FadeevSergey:
#324 
Added analysis of the following class-scoped:
* `stmt.PropertyList:`
for example
> `expr.ClassConstFetch`
> `expr.Array`
> `expr.Ternary`

* `stmt.ClassConstList:`
for example
> `expr.ClassConstFetch`
> `expr.Array`
> `expr.Ternary`

* Arguments of methods and constructors

Example: 

```
class T
{
    const C = 1;

    const C1 = self::UNDEFCONST;              // undefined class constant
    const C2 = self::C ? "var" : "var";       // then/else operands are identical

    public $var1 = self::UNDEFCONST;          // undefined class constant
    public $var2 = self::C ? "var1" : "var1"; // then/else operands are identical

    public $publicArray = [
        'key1' => 'somethingInPublic',
        'key2' => 'other_thingInPublic',
        'key1' => 'third_thingInPublic', // duplicate key 'key1'
    ];

    const constArray = [
        'key1' => 'somethingInConst',
        'key2' => 'other_thingInConst',
        'key1' => 'third_thingInConst', // duplicate key 'key1'
    ];

    public static $staticArray = [
        'key1' => 'somethingInStatic',
        'key2' => 'other_thingInStatic',
        'key1' => 'third_thingInStatic', // duplicate key 'key1'
    ];

    public $mixArrayKeys = [
        1 => 1,
	2,
	3,
    ];

    /**
     * @return int
     */
    public function foo($array = array(1, 2, 3)): int // old array syntax
    {
        return $array[0];
    }

    public function __construct($array = array(1, 2, 3)) // old array syntax
    {}
}
```

### Behavior before
No warnings.

### Actual behavior
```
const C1 = self::UNDEFCONST; - undefined class constant
const C2 = self::C ? "var" : "var"; - then/else operands are identical

public $var1 = self::UNDEFCONST; - undefined class constant
public $var2 = self::C ? "var1" : "var1"; - then/else operands are identical

'key1' => 'third_thingInPublic', - duplicate key 'key1'
'key1' => 'third_thingInConst',  - duplicate key 'key1'
'key1' => 'third_thingInStatic', - duplicate key 'key1'

public $mixArrayKeys = [
    1 => 1,
	 2,
	 3,
]; - Mixing implicit and explicit array keys

public function foo($array = array(1, 2, 3)): int - Use of old array syntax (use short form instead)

public function __construct($array = array(1, 2, 3)) - Use of old array syntax (use short form instead)
```

These changes do not pass the old tests, as they reveal new 'maybe' in the tests, namely 
`src/linttest/testdata/idn/idn.php`

`src/linttest/testdata/idn/idn.php` - was changed

PS: This PR is not related to the test assignment for the internship.

### DIFF
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -696,6 +696,15 @@ func (d *RootWalker) lowerCaseModifier(m *node.Identifier) string {
 }
 
 func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
+	sc := meta.NewScope()
+	sc.SetInClosure(false)
+	b := &BlockWalker{
+		ctx:          &blockContext{sc: sc},
+		r:            d,
+		unusedVars:   make(map[string][]node.Node),
+		nonLocalVars: make(map[string]struct{}),
+	}
+
 	cl := d.getClass()
 
 	isStatic := false
@@ -716,6 +725,17 @@ func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
 
 	for _, pNode := range pl.Properties {
 		p := pNode.(*stmt.Property)
+
+		switch exprProperty := p.Expr.(type) {
+		case *expr.Array:
+			if exprProperty.ShortSyntax {
+				b.handleArray(exprProperty)
+			}
+		case *expr.ClassConstFetch:
+			b.handleClassConstFetch(exprProperty)
+		case *expr.Ternary:
+			b.handleTernary(exprProperty)
+		}
 
 		nm := p.Variable.Name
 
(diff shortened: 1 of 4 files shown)


src/linter/root.go
```diff
@@ -717,6 +726,17 @@ func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
 	for _, pNode := range pl.Properties {
 		p := pNode.(*stmt.Property)
 
+		switch exprProperty := p.Expr.(type) {
```
YuriyNasretdinov, line 729: Why not just do node.Walk() here too, as you did with the constant list? I can't say I like the code duplication here...

PR #508, opened 2020-06-09
Adding a check for some class fields (attempt 2)

FadeevSergey:
#324 
Added analysis of the following class-scoped:
* `stmt.PropertyList:`
for example
> `expr.ClassConstFetch`
> `expr.Array`
> `expr.Ternary`

* `stmt.ClassConstList:`
for example
> `expr.ClassConstFetch`
> `expr.Array`
> `expr.Ternary`

* Arguments of methods and constructors

Example: 

```
class T
{
    const C = 1;

    const C1 = self::UNDEFCONST;              // undefined class constant
    const C2 = self::C ? "var" : "var";       // then/else operands are identical

    public $var1 = self::UNDEFCONST;          // undefined class constant
    public $var2 = self::C ? "var1" : "var1"; // then/else operands are identical

    public $publicArray = [
        'key1' => 'somethingInPublic',
        'key2' => 'other_thingInPublic',
        'key1' => 'third_thingInPublic', // duplicate key 'key1'
    ];

    const constArray = [
        'key1' => 'somethingInConst',
        'key2' => 'other_thingInConst',
        'key1' => 'third_thingInConst', // duplicate key 'key1'
    ];

    public static $staticArray = [
        'key1' => 'somethingInStatic',
        'key2' => 'other_thingInStatic',
        'key1' => 'third_thingInStatic', // duplicate key 'key1'
    ];

    public $mixArrayKeys = [
        1 => 1,
	2,
	3,
    ];

    /**
     * @return int
     */
    public function foo($array = array(1, 2, 3)): int // old array syntax
    {
        return $array[0];
    }

    public function __construct($array = array(1, 2, 3)) // old array syntax
    {}
}
```

### Behavior before
No warnings.

### Actual behavior
```
const C1 = self::UNDEFCONST; - undefined class constant
const C2 = self::C ? "var" : "var"; - then/else operands are identical

public $var1 = self::UNDEFCONST; - undefined class constant
public $var2 = self::C ? "var1" : "var1"; - then/else operands are identical

'key1' => 'third_thingInPublic', - duplicate key 'key1'
'key1' => 'third_thingInConst',  - duplicate key 'key1'
'key1' => 'third_thingInStatic', - duplicate key 'key1'

public $mixArrayKeys = [
    1 => 1,
	 2,
	 3,
]; - Mixing implicit and explicit array keys

public function foo($array = array(1, 2, 3)): int - Use of old array syntax (use short form instead)

public function __construct($array = array(1, 2, 3)) - Use of old array syntax (use short form instead)
```

These changes do not pass the old tests, as they reveal new 'maybe' in the tests, namely 
`src/linttest/testdata/idn/golden.txt`
and
`src/linttest/testdata/underscore/golden.txt`

They have been changed

PS: This PR is not related to the test assignment for the internship.

### DIFF
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -696,6 +696,15 @@ func (d *RootWalker) lowerCaseModifier(m *node.Identifier) string {
 }
 
 func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
+	sc := meta.NewScope()
+	sc.SetInClosure(false)
+	b := &BlockWalker{
+		ctx:          &blockContext{sc: sc},
+		r:            d,
+		unusedVars:   make(map[string][]node.Node),
+		nonLocalVars: make(map[string]struct{}),
+	}
+
 	cl := d.getClass()
 
 	isStatic := false
@@ -716,6 +725,17 @@ func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
 
 	for _, pNode := range pl.Properties {
 		p := pNode.(*stmt.Property)
+
+		switch exprProperty := p.Expr.(type) {
+		case *expr.Array:
+			if exprProperty.ShortSyntax {
+				b.handleArray(exprProperty)
+			}
+		case *expr.ClassConstFetch:
+			b.handleClassConstFetch(exprProperty)
+		case *expr.Ternary:
+			b.handleTernary(exprProperty)
+		}
 
 		nm := p.Variable.Name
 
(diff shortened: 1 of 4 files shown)


quasilyte:
@YuriyNasretdinov what do you think?

This change looks fine to me. There are some problems there, for example, we're not doing a proper custom checkers initialization for block walkers, but it can easily be fixed by adding a BlockWalker constructor that can be used at least in 3 places.

The other problem is that it might be a little inefficient design to do it this way as block walkers carry more context/scope information than we probably need to traverse a simple initialization statement.

We can merge this just to detect a few more code issues and then address the fundamental issue later (as it needs more thinking and planning).

src/linter/root.go
```diff
@@ -696,6 +696,15 @@ func (d *RootWalker) lowerCaseModifier(m *node.Identifier) string {
 }
 
 func (d *RootWalker) enterPropertyList(pl *stmt.PropertyList) bool {
+	sc := meta.NewScope()
+	sc.SetInClosure(false)
```
quasilyte, line 700: I'm pretty sure a Scope is "out of closure" by default, so this call is redundant.

quasilyte:

PR #509, opened 2020-06-10
Added checks for non-literal key duplicates in dupArrayKeys.

i582:
This pull request solves #325.

Now duplicates are checked for the following:

1. Strings
2. Integers
3. Real numbers (PHP rounds real numbers down, so keys 14.6 and 14.5 are cast to 14 and this will be a duplication of keys)
4. Constants
5. Class constants
6. Variables
7. Access to array elements
8. Addition
9. Subtraction
10. Multiplication
11. Division
12. Concatenation

I did not add various calls to functions and methods, since if this is in the code, then such a check will not help :smile:

Duplicates are now in the expressions, however, the expressions must be **completely identical** for the duplicate to be found. An exception is expressions with two operands, where operands can be swapped. Duplicates are found only in expressions made up of elements that are described at the beginning.

In order to be able to find duplicates from expressions, I changed the method itself. Now the **array stores not the string values ​​of the keys, but the nodes that describe the key subtrees**.

This allows you to **compare two subtrees for the keys, and if they match, it means that the keys are duplicated**.

For comparison, created `compareSubTreeForArrayItemsKey ` function:

```go
func compareSubTreeForArrayItemsKey (firstSubTreeNode node.Node, secondSubTreeNode node.Node) bool
```

Which is quite easily extensible. This function is recursive and visits every node, each of the subtrees and compares them.

Due to the fact that the method was changed, to write to the console that part of the warning that describes the value of the duplicate node, `writer` was used, which led to the fact that there could be a duplicate single quote, as in the example in` basic_test.go` :

```
"Duplicate array key ''key1''"
```

---
This is done as part of a VK internship.


quasilyte:
Hello @i582 !

Here is the agenda:

i582:
**Update n.2**

Now, the linter detects duplicates in expressions consisting of:

1. BitwiseAnd
2. BitwiseOr
3. BitwiseXor
4. Concat
5. Plus
6. Minus
7. Mul
8. Div
9. Mod
10. ArrayDimFetch
11. ClassConstFetch
12. ConstFetch
13. FunctionCall
14. MethodCall
15. StaticCall
16. StaticPropertyFetch
17. UnaryMinus
18. UnaryPlus
19. SimpleVar
20. Dnumber
21. Lnumber
22. String

These are the main operands and operators that can be used in expressions for an array key.

Now the expressions **do not have to be completely identical!**, for example, if the expressions differ only in the order of the operands (and the operator is commutative), then a duplicate will be found.

Generally speaking, **if two expressions give the same result, they will be found and marked as duplicates.**


The following set of examples describes which duplicates the linter now finds:

```php

const B1 = 10;
const B2 = 100;

class Boo {
	const B1 = 10;
	const B2 = 100;

	/**
	* @return int
	*/
	public function someFunction() {
		return 100;
	}

	/**
	* @return int
	*/
	static function f() {
		return 5;
	}
}

// concat
$example0 = [
    "Hello " . "World" => 1,
    "Hello " . "World" => 2 // Duplicate key "Hello " . "World"
];

// binary + and -
$example1 = [
    10 + 5 - 3 => 1,
    5 - 3 + 10 => 2 // Duplicate key 5 - 3 + 10
];

// binary * and / and %
$example2 = [
    10 * 20 / 4  => 1,
    50 => 2, // Duplicate key 50
    20 => 3,
    420 % 400 => 2 // Duplicate key 420 % 400
];

// simple var
$someValue = 2;
$example3 = [
    $someValue - 3 + 5 => 1,
    2 + $someValue => 2 // Duplicate key 2 + $someValue
];

// bitwise AND
$example4 = [
  0b1111 & 0b0010 => 1,
  0b0010 => 2, // Duplicate key 0b0010
];

// bitwise OR
$example5 = [
  0b1001 | 0b0010 => 1,
  0b1011 | 0b0000 => 2, // Duplicate key 0b1011 | 0b0000
];

// unary minus
$someValue = 2;
$example6 = [
  -$someValue => 1,
  -1 * $someValue => 2, // Duplicate key  -1 * $someValue
];

// function call
function foo()
{
    return 5;
}

$example7 = [
  foo() | 0b1011 => 1,
  foo() | 0b1011 | 0b0000 => 2, // Duplicate key  foo() | 0b1011 | 0b0000
];

// array access
$numbers = [10, 5, 3];

$example8 = [
  $numbers[0b1001 | 0b0010] => 1,
  $numbers[0b1011 | 0b0000] => 2, // Duplicate key $numbers[0b1011 | 0b0000]
];

// class const fetch
$example9 = [
  Boo::B1 * 10 => 1,
  Boo::B1 * 5 * 2 => 2, // Duplicate key Boo::B1 * 5 * 2
];

// const fetch
$example10 = [
  B1 * 10 => 1,
  B1 * 5 * 2 => 2, // Duplicate key B1 * 5 * 2
];

// method call

$booo = new Boo();

$example11 = [
  -$booo->someFunction() + 10 => 1,
  10 - $booo->someFunction() => 2, // Duplicate key 10 - $booo->someFunction()
];


// static call

$example12 = [
 10 / 5 + Boo::f() => 1,
  2 + Boo::f() => 2, // Duplicate key 2 + Boo::f()
];


// static property fetch
class BooMore {
    static $value = 10;
}

$example13 = [
  BooMore::$value => 1,
  BooMore::$value => 2, // Duplicate key BooMore::$value
];

// real number
// Note: PHP rounds down real numbers in keys, therefore,
// keys 14.6 and 14.5 will be one key equal to 14.
$example14 = [
  14.5 => 1,
  14.6 => 2, // Duplicate key 14.6
];

```

Thus, now, the number of different variations that the linter can detect is increasing.

The essence of the method is to calculate the hash for each of the subtrees. If the received hashes match, then both expressions describe the same expressions (that is, their result will be the same). To calculate the hash, the `getHashForExpressionNode` function is used.

The hash calculation function works very quickly, so performance will not drop because of it.

i582:
**Update n.5**

1. Improved search algorithm, complexity changed from O(n^2) to O(n).

2. Fixed a bug in a situation where the keys contained expressions that were not processed into a hash function, and because of this, false positives occurred.

3. Fixed a bug in the processing of numbers in different number systems.

4. Added processing of 'PropertyFetch'.

5. Removed helper functions.

6. Added new tests.

i582:
Closed until better times 😃

PR #510, opened 2020-06-10
Add handling non-literal array keys

alyoanton9:
Solving #325 . Now `handleArrayItems` handles not only `Lnumber` and `String`, but `ConstFetch`, `ClassConstFetch` (processed individually in case of having convenient functions to deal with namespaces) and a lot of other legal keys that satisfy `sideEffectFree` (processed by `printer`).
I suppose debatable cases like: `$example = [new T() => 1, new T() => 2,];`
should be reported by some "IllegalArrayKeys", so my `isLegalKey` function returns false if key type is `expr.New` or `expr.Array` or other illegal type (suppose this function isn't ideal and I missed some cases or included extra, but all reasonable tests were passed).
One more issue is the fact I understand we can't determine variable (like `node.Var`) type when it's array key. So "dupArrayKeys" will find duplicate `$k` also if it's array: `$k = [1, 2, 3];`. I think it can be fixed somewhere in `parser`, but it's a task for "IllegalArrayKeys". But I also can be mistaken whether linter needs this check.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1275,6 +1275,25 @@ func (b *BlockWalker) handleArrayItems(arr node.Node, items []*expr.ArrayItem) b
 		case *scalar.Lnumber:
 			key = k.Value
 			constKey = true
+		case *scalar.Dnumber:
+			key = k.Value
+			constKey = true
+		case *expr.ConstFetch:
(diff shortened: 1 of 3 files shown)


quasilyte:

PR #511, opened 2020-06-14
fix for dupplicate array keys assertion

sergalb:
solution for #325 
* add assert for node, satisfied sideEffectsFree()
* improve assert for numbers in different formats


quasilyte:
Hello @sergalb !

Here is the agenda:

quasilyte:

Issue #406, opened 2020-04-12
Report non-public interface methods


quasilyte:
## Code Example

```php
interface A {
  public function f(); // OK.
}

interface B {
  protected function f(); // Not OK.
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Report that using non-public access level in interfaces is illegal.

PR #512, opened 2020-06-15
Added report for illegal non-public interface methods and constants

i582:
#406

Added checks for non-public methods and constants in interfaces.

Example:

```php
interface WithConstants {
  const r = 10000; // ok
  public const v = 1; // ok
  private const b = 100, c1 = 1000; // 'b' can't be private, 'c1' can't be private
  protected const c = 100; // 'c' can't be protected
}

interface WithMethods {
  public function c(); // ok
  private function b(); // 'b' can't be private
  protected function f(); // 'f' can't be protected
}

interface WithStaticMethods {
  static function f1(); // ok
  public static function f2(); // ok
  private static function bad1(); // 'bad1' can't be private
  static protected function bad2(); // 'bad2' can't be protected
}

interface WithoutAnyModifier {
    function f(); // Ok,
}
```

### Behavior before
No warnings.

### Actual behavior
```
<critical> WARNING nonPublicInterfaceMember: 'b' can't be private at G:\Programming\noverify\example\index.php:19
  private const b = 100, c1 = 1000; // 'b' can't be private, 'c1' can't be private
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING nonPublicInterfaceMember: 'c1' can't be private at G:\Programming\noverify\example\index.php:19
  private const b = 100, c1 = 1000; // 'b' can't be private, 'c1' can't be private
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING nonPublicInterfaceMember: 'c' can't be protected at G:\Programming\noverify\example\index.php:20
  protected const c = 100; // 'c' can't be protected
  ^^^^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING nonPublicInterfaceMember: 'b' can't be private at G:\Programming\noverify\example\index.php:25
  private function b(); // 'b' can't be private
  ^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING nonPublicInterfaceMember: 'f' can't be protected at G:\Programming\noverify\example\index.php:26
  protected function f(); // 'f' can't be protected
  ^^^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING nonPublicInterfaceMember: 'bad1' can't be private at G:\Programming\noverify\example\index.php:32
  private static function bad1(); // 'bad1' can't be private
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING nonPublicInterfaceMember: 'bad2' can't be protected at G:\Programming\noverify\example\index.php:33
  static protected function bad2(); // 'bad2' can't be protected
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -503,7 +503,7 @@ func (b *BlockWalker) handleInterface(int *stmt.Interface) bool {
 		switch x := st.(type) {
 		case *stmt.ClassMethod:
 			for _, modifier := range x.Modifiers {
-				if modifier.Value != "public" {
+				if strings.ToLower(modifier.Value) != "public" {
 					interfaceName := int.InterfaceName.Value
 					methodName := x.MethodName.Value
 					b.r.Report(x, LevelWarning, "nonPublicInterfaceMethod", "Non-public method '%s' in the interface '%s'", methodName, interfaceName)
@@ -511,7 +511,7 @@ func (b *BlockWalker) handleInterface(int *stmt.Interface) bool {
 			}
 		case *stmt.ClassConstList:
 			for _, modifier := range x.Modifiers {
-				if modifier.Value != "public" {
+				if strings.ToLower(modifier.Value) != "public" {
 					interfaceName := int.InterfaceName.Value
 					for _, constant := range x.Consts {
 						constantName := constant.(*stmt.Constant).ConstantName.Value
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -503,26 +503,24 @@ func (b *BlockWalker) handleInterface(int *stmt.Interface) bool {
 		switch x := st.(type) {
 		case *stmt.ClassMethod:
 			for _, modifier := range x.Modifiers {
-				if strings.ToLower(modifier.Value) != "public" {
-					interfaceName := int.InterfaceName.Value
+				if strings.EqualFold(modifier.Value, "private") || strings.EqualFold(modifier.Value, "protected") {
 					methodName := x.MethodName.Value
-					b.r.Report(x, LevelWarning, "nonPublicInterfaceMethod", "Non-public method '%s' in the interface '%s'", methodName, interfaceName)
+					b.r.Report(x, LevelWarning, "nonPublicInterfaceMember", "'%s' can't be %s", methodName, modifier.Value)
 				}
 			}
 		case *stmt.ClassConstList:
 			for _, modifier := range x.Modifiers {
-				if strings.ToLower(modifier.Value) != "public" {
-					interfaceName := int.InterfaceName.Value
+				if strings.EqualFold(modifier.Value, "private") || strings.EqualFold(modifier.Value, "protected") {
 					for _, constant := range x.Consts {
 						constantName := constant.(*stmt.Constant).ConstantName.Value
-						b.r.Report(x, LevelWarning, "nonPublicInterfaceMethod", "Non-public constant '%s' in the interface '%s'", constantName, interfaceName)
+						b.r.Report(x, LevelWarning, "nonPublicInterfaceMember", "'%s' can't be %s", constantName, modifier.Value)
 					}
 				}
 			}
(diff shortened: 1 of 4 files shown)


src/linter/block.go
```diff
@@ -438,9 +438,7 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
 			res = false
 		}
 	case *stmt.Interface:
-		if b.ignoreFunctionBodies {
-			res = false
-		}
+		res = b.handleInterface(s)
```
quasilyte, line 441: `handleInterface` always returns `false`.
I believe it should return `!b.ignoreFunctionBodies`.

Otherwise, the block walker will check function bodies during the top-level traversal.

src/linttest/regression_test.go
```diff
@@ -1243,3 +1243,31 @@ function f($x) {
 }
 `)
 }
+
+func TestIssue406(t *testing.T) {
```
quasilyte, line 1247: Regression tests are mostly for fixes of the old code.
New checks are considered to be a new feature and their tests usually go in other files (almost any file will do).

The main purpose of regression test is that even if it looks like it's excessive, it should never be removed as it's bound to the bug fix. Contrary to that, tests from other files can be removed if they become unnecessary.

src/linttest/regression_test.go
```diff
@@ -1243,3 +1243,31 @@ function f($x) {
...
+	test := linttest.NewSuite(t)
+	test.AddFile(`<?php
+
+interface A {
```
quasilyte, line 1251: Another test case idea: method and constant without explicit access modifier.
PHP implies `public` for these.
These cases should not give a warning (we may give a warning about missing access modifier later in a separate check, so it can be disabled separately).

src/linter/block.go
```diff
@@ -500,6 +498,33 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
 	return res
 }
 
+func (b *BlockWalker) handleInterface(int *stmt.Interface) bool {
+	for _, st := range int.Stmts {
+		switch x := st.(type) {
+		case *stmt.ClassMethod:
+			for _, modifier := range x.Modifiers {
+				if strings.ToLower(modifier.Value) != "public" {
```
quasilyte, line 506: It's possible to use `strings.EqualFold(modifier.Value, "public")` to do a case insensitive string equality check that is usually more efficient (https://golang.org/pkg/strings/#EqualFold).

src/linter/block.go
```diff
@@ -500,6 +498,33 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
...
+					interfaceName := int.InterfaceName.Value
+					methodName := x.MethodName.Value
+					b.r.Report(x, LevelWarning, "nonPublicInterfaceMethod", "Non-public method '%s' in the interface '%s'", methodName, interfaceName)
```
quasilyte, line 509: This diagnostic name in combination with the warning text will result in a stutter:
```
nonPublicInterfaceMethod: Non-public constant 'b' in the interface 'A'
```
It's better to avoid that. We say "non public in interface" twice on the same line.
It may be clearer this way:
```
nonPublicInterfaceMethod: 'b' can't be private
```
The diagnostic name reads like a message prefix, so it's redundant to repeat it.

I also suggest giving both warnings as a single diagnostic, `nonPublicInterfaceMember`.

src/linter/block.go
quasilyte: I also think that it will give a false positive for `static` methods.

Add this test case to be sure:
```php
interface WithStatic {
  static function f1(); // OK
  public static function f2(); // OK
  private static function bad1(); // Bad: private
  static protected function bad2(); // Bad: protected
}
```

i582:
Added fixes

PR #514, opened 2020-06-18
Handling non-literal key duplicates in dupArrayKeys diagnostic

qypec:
**Solving #325**

Handling for the most nodes without side effect.
Tests for cases with this nodes.

Tracked various numeral systems for `Lnumber`:
```
6464 => 1,
014500 => 2, // Duplicate key '6464'
0x1940 => 3, // Duplicate key '6464'
```

Illegal types:
- `expr.New`
- `expr.FunctionCall`

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -3,6 +3,7 @@ package linter
 import (
 	"fmt"
 	"regexp"
+	"strconv"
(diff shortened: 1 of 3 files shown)


quasilyte:

PR #515, opened 2020-06-18
Solving issue #325

FadeevSergey:
#325
Added diagnostics duplication of non-literal keys added. Now the following keys are checked:
```
scalar.Dnumber,
node.SimpleVar,
expr.ConstFetch,
binary.Concat,
expr.ClassConstFetch,
expr.PropertyFetch,
expr.ArrayDimFetch,
expr.StaticPropertyFetch,
expr.Ternary,
expr.FunctionCall,
expr.MethodCall,
expr.StaticCall,
expr.InstanceOf 
```
and everyone `node.Node` that satisfying `sideEffectFree`.

Improved node.Lnumber checking. Now the same numbers in different number systems are considered the same. For example: 
```
$array = [
    1234 => 7,
    02322 => 8,  // 02322 = 1234 duplicate
    0x4D2 => 9,  // 0x4D2 = 1234 duplicate
    0b10011010010 => 10,// 0b10011010010 = 1234 duplicate
    1_234 => 11, // 1_234 = 1234 duplicate
];
```
Keys are compared by their identifier. Key identifier is looking for getNodeIdentifire. getNodeIdentifier returns a string representation of the key.
* All scalar.Strings are recorded with apostrophes.
Note that in the following example, $ str is of type not scalar.String:
```
$var = 1;
$str = "$var";
```
* All scalar.Lnumber are converted to decimal number system.  
* All scalar.Dnumber translates to a record with 15 characters after the decimal point.  
* The last comma in the key-value record is deleted (for example, [1, 2, 3,] is converted to [1, 2, 3]). 
 
The above replacements are made in:
1) `scalar.String`
2) `scalar.Lnumber`
3) `scalar.Dnumber`
4) `binary.Concat`
5) `expr.Ternary`
6) `expr.FunctionCall`
7) `expr.MethodCall`
8) `expr.StaticCall`

In other cases, a string representation of the key is returned, in which extra spaces are removed and the necessary ones are added.
For example, the following keys will be considered the same:
```
* $str      .   $classT->str .    $k[$classT->var] . $k[$k[$k[$classT->nameOfClassA::C1]]] . T::staticFunc(...[1 => 1, 2 => 2])
* $str.$classT->str.$k[$classT->var].$k[$k[$k[$classT->nameOfClassA::C1]]].T::staticFunc( ...[1 => 1,    2 => 2, ])
```

In the future, changing strings and numbers can be added for all binary operations. Keys with only commutative operations can be compared not by their string representation, but by the node on which the operation is performed.
This PR is the solution to the internship assignment. Better yet compare AST

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -363,7 +363,6 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
 	case *stmt.Try:
 		res = b.handleTry(s)
 	case *assign.Assign:
-		//fmt.Println(s.Position)
 		// TODO: only accept first assignment, not all of them
 		// e.g. if there is a condition like ($a = 10) || ($b = 5)
 		// we must only accept $a = 10 as condition that is always executed
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -3,6 +3,7 @@ package linter
 import (
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/VKCOM/noverify/src/meta"
@@ -362,6 +363,7 @@ func (b *BlockWalker) EnterNode(w walker.Walkable) (res bool) {
 	case *stmt.Try:
 		res = b.handleTry(s)
 	case *assign.Assign:
+		//fmt.Println(s.Position)
 		// TODO: only accept first assignment, not all of them
 		// e.g. if there is a condition like ($a = 10) || ($b = 5)
 		// we must only accept $a = 10 as condition that is always executed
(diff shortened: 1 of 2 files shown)


FadeevSergey:
Comment has been changed

FadeevSergey:
1) Need to delete line 1332 in block.go
2) On line 1377, `astutil.FmtNode(k.Variable)` should be replaced with `b.getNodeIdentifier(k.Variable)`. This will allow additional duplicate keys to be detected.

quasilyte:

PR #517, opened 2020-06-19
Partial solution of the duplicate array keys issue 325

d0rj:
Partially fixed the problem. Duplicates of array keys can now be: numbers in any format, strings, variables, constants, return values of functions, etc. Also, the `dupArrayKeys` does not check situations that are not related to its problem.

However, more complex expressions, such as binary operations, are not checked:
```php
$arr = [
    -($a + $b) => 1,
    -$b - $a => 2,
];
```
and many others, in which the use of nodes included in the expression differs only in order.

The solution to this, it seems to me, cannot be reduced to calculating the value of a number or to comparing a string representation. But the solution, I think, is a comparison of AST or the use of standardized reduction of term strings. For example:
```go
"-($b + $a)" ~> "-$b - $a" ~> "-$a - $b" ~> "-$a-$b"
```
It seems to me that the use of memory and computing resources in this case will not be too critical, but still greater than at the moment.

_Test task for admission to the VK internship._

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1275,7 +1275,7 @@ func (b *BlockWalker) handleArrayItems(arr node.Node, items []*expr.ArrayItem) b
 		switch k := item.Key.(type) {
 		case *scalar.String:
 			keyString = unquote(k.Value)
-			warningKeyString = k.Value
+			warningKeyString = keyString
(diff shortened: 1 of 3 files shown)


i582:
Sorry for that, this is a mistake.

PR #518, opened 2020-06-20
Fix issue 325: report dupArrayKeys

K1ta:
To detect duplicates I cast all nodes to string names, like php does. E.g. if node type is string, first I check if string begins with '0' or '+', if not, I try to parse it to int, else I save it in keys map with quotes. 

Duplicate array keys detection now works for:
* numbers (integer and float)
* strings
* booleans
* constants and class-constants
* simple variables
* array elements

Some things to improve:
* check if array from last item is not a class that implements ArrayAccess
* simplify handling Lnumber and Dnumber, UnaryMinus and UnaryPlus

Also I noticed, that this code:
```php
<?php 
$test1 = [true => 2];
?>
```
causes error: `ERROR undefined: Use true instead of true at _file0.php:2`. Is this expected behaviour? I can not write test for boolean keys duplicates because of this error, it is commented now.


quasilyte:

PR #519, opened 2020-06-21
Non-literal key duplicates in dupArrayKeys diagnostic

Dmitrii-Stukalov:
Solving #325 
Improve handling ints (different numeral systems)
Add handling constants and side effect free nodes
Add tests


quasilyte:

PR #520, opened 2020-06-21
master

zhelenskiy:


zhelenskiy:
As described in issue #325, there was no support of non-literal keys when checking for duplicates. I added such checks and improved warning messages. They show the place of the first occurrence and the current one now too.

Supported:
* expr.ConstFetch
* expr.ClassConstFetch
* scalar.Lnumber _(invariant format)_
* scalar.Dnumber _(invariant format)_
* scalar.String _(even with heredocs)_
...
* Any pure functions

Each part of the code is covered by tests.

quasilyte:

PR #522, opened 2020-06-23
Added checks for constant, property and method duplicates in classes

i582:
Added warnings if the class contains duplicate:

1. Constants
2. Properties
3. Methods

## Example:
```php
/**
 * @property int $d Just property
 * @property int $ok Just property
 * @method static void foo() Just method
 * @method void boo() Just method
 * @method void coo() Just method
 */
class B {

public const b = 100;
private const b = 100;
protected const b = 100; // Constant B::b cannot be redefine


public $d;
public static $d = 100;
protected static $d = 100; // Property B::$d cannot be redeclare

/**
 * Some method foo()
 */
public static function foo() {}
/**
 * Some method foo()
 */
private function foo() {}
/**
 * Some method foo()
 */
protected function foo() {} // Method B::foo cannot be redeclare

/**
 * Some method boo()
 */
public function boo() {}

/**
 * Some method boo()
 */
protected function boo() {} // Method B::boo cannot be redeclare


public $ok; // All ok

/**
 * Some method coo()
 */
public function coo() {} // All ok

}
```

## Behavior before
No warnings.

## Actual Behavior:
```
<critical> ERROR   classConstantRedefinition: Constant B::b cannot be redefine at G:\Programming\noverify\example\index.php:22
private const b = 100;
              ^^^^^^^
<critical> ERROR   classConstantRedefinition: Constant B::b cannot be redefine at G:\Programming\noverify\example\index.php:23
protected const b = 100; // Constant B::b cannot be redefine
                ^^^^^^^
<critical> ERROR   classPropertyRedeclaration: Property B::$d cannot be redeclare at G:\Programming\noverify\example\index.php:27
public static $d = 100;
              ^^^^^^^^
<critical> ERROR   classPropertyRedeclaration: Property B::$d cannot be redeclare at G:\Programming\noverify\example\index.php:28
protected static $d = 100; // Property B::$d cannot be redeclare
                 ^^^^^^^^
<critical> ERROR   classMethodRedeclaration: Method B::foo cannot be redeclare at G:\Programming\noverify\example\index.php:37
private function foo() {}
^^^^^^^^^^^^^^^^^^^^^^^^^
<critical> ERROR   classMethodRedeclaration: Method B::foo cannot be redeclare at G:\Programming\noverify\example\index.php:41
protected function foo() {} // Method B::foo cannot be redeclare
^^^^^^^^^^^^^^^^^^^^^^^^^^^
<critical> ERROR   classMethodRedeclaration: Method B::boo cannot be redeclare at G:\Programming\noverify\example\index.php:52
protected function boo() {} // Method B::boo cannot be redeclare
^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

> To distinguish between properties and methods from PHPDoc and real ones, the FromDoc property of FuncInfo and PropertyInfo was added.
As I understand it, because new fields were added to the structures, it was necessary to change the cache version.

Tests were conducted on 3 projects:

1. laravel
2. yii2
3. PHPMailer

There were no false and any positives.


i582:
Based on the current changes, two new checks can be added.

1. The first will verify that all the properties and methods described in PHPDoc were actually added to the class.
2. And the second will check the opposite, that is, the presence of a description of this method or property in PHPDoc.

What do you think about this?

quasilyte:
`@method` and `@property` annotations are usually used to document "magic" members that are not explicitly declared.

In other words, if someone is using `@property`, they probably don't want to declare that property in a normal way. I would guess that it's possible to report cases where you have both a normal field definition and a `@property` annotation, but I'm not entirely sure.

`@method` is used for magic methods that can be implemented via `__call` or `__callStatic`.

src/linter/report.go
```diff
@@ -263,6 +263,24 @@ func init() {
...
+		{
+			Name:    "classPropertyRedeclaration",
+			Default: true,
+			Comment: `Report redeclaration of class property.`,
+		},
+
+		{
+			Name:    "classMethodRedeclaration",
+			Default: true,
+			Comment: `Report redeclaration of class method`,
```
quasilyte, line 282: I'm not sure that we need 3 different checks for this. They're checking a single logical error.

src/linter/phpdoc_util.go
```diff
@@ -68,6 +68,7 @@ func parseClassPHPDocMethod(ctx *rootContext, result *classPhpDocParseResult, pa
 		Flags:        funcFlags,
 		MinParamsCnt: 0, // TODO: parse signature and assign a proper value
 		AccessLevel:  meta.Public,
+		FromDoc:      true,
```
quasilyte, line 71: I believe it's more efficient to add another flag to a Func/Property.
I'm not sure that we have a `Flags` field for properties, but we definitely have flags for functions and classes.

src/meta/metainfo.go
```diff
@@ -353,6 +354,8 @@ type PropertyInfo struct {
 	Pos         ElementPosition
 	Typ         TypesMap
 	AccessLevel AccessLevel
+	FromDoc     bool
+	IsVerified  bool
```
quasilyte, line 358: `IsVerified` looks like a strange name if taken in comparison with `FromDoc`. It looks like it's used as `FromDecl` when the property is defined using a PHP declaration instead of `@property`.

Why do we have IsVerified only for PropertyInfo, but not for the FuncInfo?

src/linttest/oop_test.go
```diff
@@ -1061,3 +1061,84 @@ trait AbstractTraitAB {
...
+}
+`)
+	test.Expect = []string{
+		`Constant B::b cannot be redefine`,
+		`Constant B::b cannot be redefine`,
+		`Property B::$d cannot be redeclare`,
+		`Property B::$d cannot be redeclare`,
+		`Method B::foo cannot be redeclare`,
+		`Method B::foo cannot be redeclare`,
+		`Method B::boo cannot be redeclare`,
```
quasilyte, line 1141: s/redeclare/redeclared/

src/linttest/oop_test.go
```diff
@@ -1061,3 +1061,84 @@ trait AbstractTraitAB {
...
+/**
+ * Some method boo()
+ */
+protected function boo() {} // Method B::boo cannot be redeclare
+
+}
+`)
+	test.Expect = []string{
+		`Constant B::b cannot be redefine`,
+		`Constant B::b cannot be redefine`,
```
quasilyte, line 1136: s/redefine/redefined/

src/meta/metainfo.go
i582: I could not come up with a good name, but this flag is needed so that there are no false positives.

This flag is set to true only when this property has been added to functions where duplicates are checked (`enterPropertyList`) if the property in the class has already been added from somewhere outside (because there have been cases when the class already had some properties (not from documentation), but class processing has just begun), then this flag will always be false, thereby only those properties that were added to `enterPropertyList` are checked, which will help to avoid any false positives.

We can say that with this flag, the `FromDoc` flag becomes redundant, since for all properties from the documentation `IsVerified` will be false.

i582: Probably the truth is that can rename this property `FromDecl` and leave only it.

i582:
Added fixes

i582:
Closed until better times 😃

Issue #523, opened 2020-06-23
Reduce the amount of global vars in noverify


quasilyte:
This is a tracking issue for all commits that try to reduce the number of global vars we have in noverify.

i582:
Not used for a long time.

PR #524, opened 2020-06-23
cmd: replace the global flag vars with an explicit context

quasilyte:
It's a small step, but it's better to do it now than never.

The first benefit that we get is that now it's obvious that
every InitEmbeddedRules call from the linttest package also
modified cmd package global state. It's not an error, but
it's not a pleasant side effect.

The next step would be to turn all linter.* vars from the
config.go file into a struct that would them and pass it
to the root walkers.

Updates #523

### DIFF
--- a/src/cmd/flags.go
+++ b/src/cmd/flags.go
@@ -3,10 +3,8 @@ package cmd
 import (
 	"flag"
 	"fmt"
-	"io"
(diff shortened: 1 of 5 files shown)


quasilyte:
Renamed the type and its usages.

Issue #525, opened 2020-06-25
Baseline experiment


quasilyte:
- [ ] implement baseline file generation and use it to filter warnings
- [ ] test baseline approach
- [ ] update docs (if the experiment is a success)

Baseline approach overview, as well as its comparison with diff mode, can be found here: https://habr.com/ru/post/508094/ (ru).

PR #530, opened 2020-06-27
implement baseline linting mode

quasilyte:
-output-baseline is a new output mode that will produce a baseline
file instead of normal reports. This option is used to generate
a suppression profile.

-baseline option enables baseline mode and uses a provided
suppression profile to filter out the output.

-conservative-baseline flag disables hashing of the prev and next
lines of the source code line being reported. This can be useful
if we want the behavior that is closer to the original diff mode,
but it may be more prone to false negatives. We need to collect
some stats on diagnostics and see which ones of them produce
more hash collisions on the real projects (can be solved by
making their report message more detailed and dynamic).
    

Baseline file is stored in JSON format and has several
size optimizations in 3 metrics:

	1. Line width
	2. Line numbers
	3. Overall file size

(1) and (2) are important to ensure that a file is openable
in most text editors. It also makes a file more git-friendly.

Warnings are converted to a 64-bit hash using the following fields:

	* Filename (only base)
	* Previous source line
	* Start source line
	* Next soure line
	* Warning message (includes a check name)
	* Warning scope (global/func/method name)

Updates #525


quasilyte:
Added the `-conservative-baseline` flag. The description is available inside the commit message + PR first message.

quasilyte:
decoding speed (75000 reports):
```
method        elapsed  size   comment
gob           ~0.037s  1.0MB  binary format
JSON (base64) ~0.040s  1.4MB  custom strconv functions
JSON (base36) ~0.055s  1.6MB  base36 is the stdlib strconv limit
JSON (base10) ~0.064s  2.0MB  just for comparison
```
If we use gob on normal `Profile`, we'll get `2.0MB` file and x2-3 times slower decoding. All results above acquired by using `jsonProfile` that uses sorted slices instead of maps (we need stable output anyway, so maps are out of the question).

We could use base36 with a slight loss in profile compression and decoding speed (which probably happens due to the bigger file size).

quasilyte:
Collisions with -conservative-baseline:
```
944	arraySyntax "Use of old array syntax (use short form instead)"
829	redundantCast "expression already has $T type"
748	undefined
322	deprecated
228	callStatic "Calling static method as instance method"
99	caseContinue "'continue' inside switch is 'break'"
74	mixedArrayKeys "Mixing implicit and explicit array keys"
25	unused
5	voidResultUsed "void function result used"
5	discardExpr "expression evaluated but not used"
```
Messages with constant text are the most common source of collisions. It looks like at least some of them can be fixed.

Issue #513, opened 2020-06-16
Add diff mode tests


quasilyte:
noverify is only tested in normal mode where it reports all issues that were found.

The diff (git) mode is what may be the most important feature for the real use-cases.

We need to add at least some tests to a diff mode to make sure it works correctly and does not break. This is especially important since git_main is scheduled for refactoring.

I had several bug reports where diff was computed in an unexpected way. It's hard to fix these issues without having a foundation for verification (manual testing is meh).

YuriyNasretdinov:
I thought about this. The issue is that you need an external tool (git) to be able to test it. It is still possible, of course. Also it would probably be pretty hard to do a full test anyway as there are a lot of components there, e.g. a `git fetch` operation, or comparison to master branch. So, in short, the tests themselves would be pretty complex and susceptible to breakage as well as the code. Probably still better than relying on manual testing I guess, although I'm not entirely convinced.

quasilyte:
Isn't it a property of most integration tests?
They're not that simple, but they test the real system.
We can have a git repository in testdata.

If `git` is not available, this test can be skipped with `t.Skip`.

quasilyte:
It seems like my recent refactoring broke the diff mode. :(

I'm not used to do a lot of manual testing after CI is passing and when I do, it's usually the normal mode run + our bash script wrapper which not necessarily cover everything (it doesn't include full diff, for starters).

If diff mode will not be replaced by the baseline mode, it would be very useful to have some tests for it. :)

PR #532, opened 2020-06-29
cmd: fix git mode for the gitRepoComputeReportsFromCommits() case

quasilyte:
Code was assigning to a wrong field so git.Zero was not replaced
by the "master" string. This led to a panic that is quite hard to
reproduce outside of the real-world environment.

To make this mistake less likely in future, I removed fields
from the struct and store these values as local values as they're
only needed inside that function scope.

gitRef looks unused. It was unused before the refactoring as well.
Maybe it was used before but we broke something a long time ago?
This calls for another investigation.

Refs #513

### DIFF
--- a/src/cmd/flags.go
+++ b/src/cmd/flags.go
@@ -51,7 +51,7 @@ type cmdlineArguments struct {
 	gitFullDiff                bool
 	gitIncludeUntracked        bool
 	gitRepo                    string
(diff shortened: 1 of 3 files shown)


YuriyNasretdinov:
gitRef shouldn't be used anymore but is left for compatibility.

PR #535, opened 2020-06-30
solver: improve `self` handling in traits

### LINKED ISSUE #533: Improve self type handling in traits
## Code Example

```php
<?php

trait Singleton {
  /**
   * @var self
   */
  private static $instance = null;

  /**
   * @return self
   */
  public static function instance() {
    if (!self::$instance) {
      self::$instance = new self();
    }

    return self::$instance;
  }
}

class Foo {
  use Singleton;
  public function f() {}
}

Foo::instance()->f();
```

## Actual Behavior

Undefined method error. `Foo::instance()` result type is resolved into `\Singleton|null`.

## Expected Behavior

No undefined method error. `Foo::instance()` result type contains `\Foo`.

Note that if we change `self` to `static` in `Singleton` definition, linter will behave as expected. This issue is about `self` type annotation that should behave identically in this case (although some manual/docs reading is suggested).

quasilyte:
When `self` is used inside a trait T, we expand that `self` to T
during the indexing phase (so metadata contains T instead of `self`).

This is not correct as T can't be a valid type on its own and `self`
always refers to a class C that embeds T.

Since T is not a valid type name, but it's known to be a trait name,
we can safely replace T with C during the types solving.

In contrast to the late static bindings resolution, we do that
replacement right away using the current scope className.

I also fixed a bug in WStaticPropertyFetch case. It was passing
class variable instead of className down the resolution chain
which led to invalid late static binding results.

Fixes #533

### DIFF
--- a/src/solver/solver.go
+++ b/src/solver/solver.go
@@ -150,10 +150,17 @@ func (r *resolver) resolveTypeNoLateStaticBinding(class, typ string) map[string]
 		className, methodName := meta.UnwrapStaticMethodCall(typ)
 		m, ok := FindMethod(className, methodName)
 		if ok {
-			return r.resolveTypes(className, m.Info.Typ)
+			res = r.resolveTypes(className, m.Info.Typ)
+			if m.TraitName != "" {
+				res = replaceTraitName(res, m.TraitName, m.ClassName)
+			}
+			return res
 		}
 		m, ok = FindMethod(className, "__callStatic")
 		if ok {
+			// Should probably run replaceTraitName here on the result
+			// as well, but I don't have a good __callStatic trait method
+			// example, so I hesitate.
 			return r.resolveTypes(className, m.Info.Typ)
 		}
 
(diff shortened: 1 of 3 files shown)


src/solver/solver.go
```diff
@@ -150,18 +150,29 @@ func (r *resolver) resolveTypeNoLateStaticBinding(class, typ string) map[string]
...
+			res = r.resolveTypes(className, m.Info.Typ)
+			if m.TraitName != "" {
+				res = replaceTraitName(res, m.TraitName, m.ClassName)
+			}
+			return res
 		}
 		m, ok = FindMethod(className, "__callStatic")
 		if ok {
+			// Should probably run replaceTraitName here on the result
+			// as well, but I have a good __callStatic trait method
```
YuriyNasretdinov, line 162: I assume you meant that you don't have a good example, not that you do?

src/solver/solver.go
```diff
@@ -570,3 +581,18 @@ func getClassOrTrait(typeName string) (meta.ClassInfo, bool) {
 	}
 	return class, false
 }
+
+// replaceTraitName replaces traitName with className inside res.
+//
+// Note: since we don't have a proper class context in non-static method
+// calls as well as other instance-related operations, we can't use this
+// outside of the static context.
```
YuriyNasretdinov, line 589: I don't really see why we couldn't do the same in dynamic contexts though. We always have a class name and a trait name when calling a method, right? It should require a more sophisticated approach than the one used here but I think it should still be possible as the end result should be the same: we replace a resulting trait name with the class that used that trait last.

quasilyte: I removed that comment part.

src/solver/solver.go
quasilyte: You're right. I fixed a typo.

src/solver/solver.go
quasilyte: Let's revisit non-static contexts after the need arises. It should be doable, as you said.

PR #550, opened 2020-07-03
linter: fix function name resolving in side effects finder

### LINKED ISSUE #547: Using builtin funcs via FQN results in unexpected warnings
## Code Example

```php
<?php

namespace foo;

\setlocale(LC_ALL, 'nl_NL');
\putenv("FOO=1");
```

## Actual Behavior

```
<critical> WARNING discardExpr: expression evaluated but not used at example.php:5
\setlocale(LC_ALL, 'nl_NL');
^^^^^^^^^^^^^^^^^^^^^^^^^^^
<critical> WARNING discardExpr: expression evaluated but not used at example.php:6
\putenv("FOO=1");
^^^^^^^^^^^^^^^^
```

## Expected Behavior

No warnings.

quasilyte:
Since we can get both FQN and relative names, it's more
reliable to do a proper function name resolving in
side effects finder.

We only handled relative names before which led to false
positives in discardExpr when FQN was used.

Fixes #547

### DIFF
--- a/src/linter/side_effects_finder.go
+++ b/src/linter/side_effects_finder.go
@@ -99,12 +99,22 @@ var pureBuiltins = func() map[string]struct{} {
 }()
 
 func (f *sideEffectsFinder) functionCallIsPure(n *expr.FunctionCall) bool {
-	// Allow referencing builtin funcs even before indexing is completed.
-	nm, ok := n.Function.(*name.Name)
+	// We allow referencing builtin funcs even before indexing is completed.
+
 	var funcName string
-	if ok && len(nm.Parts) == 1 {
+	switch nm := n.Function.(type) {
+	case *name.Name:
+		if len(nm.Parts) == 1 {
+			funcName = `\` + meta.NameToString(nm)
+		}
+	case *name.FullyQualified:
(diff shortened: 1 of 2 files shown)


src/linter/side_effects_finder.go
```diff
@@ -99,12 +100,11 @@ var pureBuiltins = func() map[string]struct{} {
 }()
 
 func (f *sideEffectsFinder) functionCallIsPure(n *expr.FunctionCall) bool {
-	// Allow referencing builtin funcs even before indexing is completed.
-	nm, ok := n.Function.(*name.Name)
-	var funcName string
-	if ok && len(nm.Parts) == 1 {
+	// We allow referencing builtin funcs even before indexing is completed.
+
+	funcName, ok := solver.GetFuncName(f.st, n.Function)
```
YuriyNasretdinov, line 105: Won't it cause a race condition? You are updating the index during indexing, you can't use it to resolve anything yet.

quasilyte: You're right.

quasilyte: Fixed.

PR #552, opened 2020-07-04
linter: don't stop at the first method/prop match

### LINKED ISSUE #548: accessLevel false positive when field name is not unique
## Code Example

```php
<?php

class A {
  private $value;
}

class B {
  private $value;

  /**
   * @param A|B $x
   */
  private function foo($x) {
    if ($x instanceof B) {
      echo $x->value;
    }
  }
}
```

## Actual Behavior

```
<critical> ERROR   accessLevel: Cannot access private property \A->value at example.php:15
      echo $x->value;
               ^^^^^
```

## Expected Behavior

No warnings.

quasilyte:
When doing a method/property lookup to see whether it's defined,
don't stop at the first name match. We may find a more accessible
version of the symbol. Right now we prioritize own symbols over
base class symbols over everything else.

Fixes #548

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -2,6 +2,7 @@ package linter
 
 import (
 	"fmt"
+	"math"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


quasilyte:
Maybe solver's FindMethod and FindProperty should take __call and __get into account?

PR #574, opened 2020-07-13
baseline: store creation date inside the profile

quasilyte:
This makes it easier to figure out which profile is older
or newer (especially if the linter version string is identical).


quasilyte:
Changed to UNIX time.

Issue #569, opened 2020-07-10
Report "you are not allowed to disable linter" only once per file


quasilyte:
This output seems excessive:
```
You are not allowed to disable linter for file '/home/robert-martin/www/the_cleanest.php'
You are not allowed to disable linter for file '/home/robert-martin/www/the_cleanest.php'
You are not allowed to disable linter for file '/home/robert-martin/www/the_cleanest.php'
You are not allowed to disable linter for file '/home/robert-martin/www/the_cleanest.php'
You are not allowed to disable linter for file '/home/robert-martin/www/the_cleanest.php'
...
```
We should probably memorize which files already had this warning and do not report it repeatedly.

PR #578, opened 2020-07-14
Solve "you are not allowed to disable linter" only once per file #569

i582:
solve #569

### DIFF
--- a/src/linter/report.go
+++ b/src/linter/report.go
@@ -263,6 +263,12 @@ func init() {
 			Default: true,
 			Comment: `Report illegal non-public access level in interfaces.`,
 		},
+
+		{
+			Name:    "linterError",
+			Default: true,
+			Comment: `Report linter error.`,
+		},
 	}
 
 	for _, info := range allChecks {
@@ -332,7 +338,11 @@ func (r *Report) String() string {
 
 	// No context line for security-level warnings.
 	if r.level == LevelSecurity {
-		return fmt.Sprintf("%s %s at %s:%d", severityNames[r.level], msg, r.filename, r.startLine)
+
+		// To make output stable between platforms, see #572
+		filename := strings.ReplaceAll(r.filename, "\\", "/")
+
+		return fmt.Sprintf("%s %s at %s:%d", severityNames[r.level], msg, filename, r.startLine)
 	}
 
 	contextLn := strings.Builder{}
--- a/src/linter/report.go
+++ b/src/linter/report.go
@@ -22,138 +22,243 @@ func init() {
 			Name:    "discardExpr",
 			Default: true,
 			Comment: `Report expressions that are evaluated but not used.`,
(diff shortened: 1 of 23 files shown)


i582:
> I would suggest to print the diagnostic at the phpdoc that tries to disable linter instead. It would be more relevant there and you don't need a map too.

If I understand correctly, do you suggest printing all places where there was an attempt to disable linter?

src/linter/root.go
```diff
@@ -568,14 +569,19 @@ func (d *RootWalker) handleComment(c freefloating.String) {
 		return
 	}
 
+	if d.handledFiles == nil {
+		d.handledFiles = make(map[string]bool)
```
quasilyte, line 573: As @YuriyNasretdinov already mentioned, you don't need a map here. Root walker is file-scoped. It doesn't handle multiple files. Having a bool field should be enough.

src/linter/root.go
```diff
@@ -568,14 +569,19 @@ func (d *RootWalker) handleComment(c freefloating.String) {
...
 	for _, ln := range phpdoc.Parse(d.ctx.phpdocTypeParser, str) {
 		if ln.Name() != "linter" {
 			continue
 		}
 
 		for _, p := range ln.(*phpdoc.RawCommentPart).Params {
-			if p == "disable" {
-				d.disabledFlag = true
+			if p == "disable" && !d.isDisabled && !d.handledFiles[d.ctx.st.CurrentFile] {
+				d.Report(n, LevelSecurity, "linterError", "You are not allowed to disable linter")
```
quasilyte, line 583: I think it should be LevelInformation or something, but definitely not LevelSecurity. Security is for very specific kinds of checks that detect potential security vulnerabilities. Using a comment where it shouldn't be used is not a security issue. :)

src/linter/report.go
```diff
@@ -332,7 +338,11 @@ func (r *Report) String() string {
 
 	// No context line for security-level warnings.
 	if r.level == LevelSecurity {
-		return fmt.Sprintf("%s %s at %s:%d", severityNames[r.level], msg, r.filename, r.startLine)
+
+		// To make output stable between platforms, see #572
```
quasilyte, line 342: This is a part of the other change list. Have you pulled the master?

src/linter/root.go
```diff
@@ -63,7 +63,8 @@ type RootWalker struct {
 
 	currentClassNode node.Node
 
-	disabledFlag bool // user-defined flag that file should not be linted
+	isDisabled   bool            // user-defined flag that file should not be linted
```
quasilyte, line 66: This should probably be called `allowDisabled` as it's not disabled if there were not `linter disable` comments inside a file. Also, I would store a regexp here as you pass it into parse contents function anyway. Then we can try to match a regexp only if we encounter the `@linter disable` comment (which is rare) instead of matching every filename against it.

Issue #583, opened 2020-07-15
Add storage of constant value to ConstantInfo


i582:
Saving the values of constants will allow you to find duplicate keys in the array if they are represented by constants with the same value.

```php
const MAX_VALUE = 1;
const MIN_VALUE = 1;

$arr = [
    MAX_VALUE => 5,
    MIN_VALUE => 6
];
```

It will also allow in such examples:
```php
const MAX_VALUE = 1;
const MIN_VALUE = 1;

if (MIN_VALUE < MAX_VALUE) {
    $a = 100;
}
```
find duplicate operand errors.

PR #584, opened 2020-07-15
linter: Added saving of constant value in ConstantInfo #583

i582:
solve #583


src/linter/block.go
```diff
@@ -1361,6 +1362,27 @@ func (b *BlockWalker) handleArrayItems(arr node.Node, items []*expr.ArrayItem) b
...
+
+			key = info.Value.Value
+
+			if info.Value.Type == meta.Float {
+				val, err := strconv.ParseFloat(key, 64)
+				if err != nil {
+					continue
+				}
+				value := int64(math.Floor(val))
+				key = fmt.Sprint(value)
```
quasilyte, line 1381: You might want to test this code path.

From the PHP manual:
> Floats are also cast to integers, which means that the fractional part will be truncated. E.g. the key 8.7 will actually be stored under 8. 

This gives a hint that values are "truncated", not floor'ed.

For some inputs, it behaves the same, but:
```
int(floor(-1.51)) => int(-2.0) => -2
int(-1.51) => -1
```
PHP will do the latter:
```php
var_dump([-1.5 => 1]);
```
```
array(1) {
  [-1]=>
  int(1)
}
```
I would suggest skipping the floor step.
It's also a good idea to add tests that check this corner case behavior for `-1.51`.

src/linter/block.go
```diff
@@ -1361,6 +1362,27 @@ func (b *BlockWalker) handleArrayItems(arr node.Node, items []*expr.ArrayItem) b
...
+			} else if info.Value.Type == meta.Undefined {
```
quasilyte, line 1382: I would suggest moving this check closer to the !ok, so we get guarding clauses above the code that handles correct known values.

src/linter/block.go
```diff
@@ -210,7 +211,7 @@ func (b *BlockWalker) checkBinaryDupArgs(n, left, right node.Node) {
 	if !b.sideEffectFree(left) || !b.sideEffectFree(right) {
 		return
 	}
-	if astutil.NodeEqual(left, right) {
+	if b.NodeEqual(left, right) {
```
quasilyte, line 214: This may now match `Foo == Bar` as duplicated if their values are identical.
It might look like a false positive for a user.
Perhaps we need to adjust the report message so it looks less broken for this kind of situation.

src/linter/root.go
```diff
@@ -857,11 +857,14 @@ func (d *RootWalker) enterClassConstList(s *stmt.ClassConstList) bool {
 		d.checkCommentMisspellings(c, c.PhpDocComment)
 		typ := solver.ExprTypeLocal(d.scope(), d.ctx.st, c.Expr)
 
+		value, _ := solver.GetConstantValue(c, d.ctx.st)
```
quasilyte, line 860: GetConstantValue may call GetConstant and this creates a bad situation where the order of the file traversal matters. We try to avoid that.

In general, you can't use solver methods that rely on the metadata during the indexing.

I would limit GetConstantValue to AST nodes and avoid querying the metadata during the indexing.

quasilyte:
I would also suggest trying to write more detailed commit messages.

This might be helpful:
https://golang.org/doc/contribute.html#commit_messages

Issue #585, opened 2020-07-15
Add the ability to create reports with a context line that displays part of PHPDoc


i582:
For example, now this may be necessary at least for an error with when `/** @linter disable */` is not allowed (Now there is an empty line in the report).

PR #587, opened 2020-07-15
ReportByLine for report with a context line with specific line

i582:
Solve #585


src/linter/root.go
```diff
@@ -484,6 +484,76 @@ func (d *RootWalker) Report(n node.Node, level int, checkName, msg string, args
 	}
 }
 
+// ReportByLine registers a single report message about some found problem in lineIndex code line.
+func (d *RootWalker) ReportByLine(lineIndex int, level int, checkName, msg string, args ...interface{}) {
```
quasilyte, line 488: Can we make a generalized report method that can implement major logic so we don't duplicate the code that manipulates that hash computations, etc?

src/linter/root.go
```diff
@@ -666,8 +624,9 @@ func (d *RootWalker) handleComment(c freefloating.String) {
 			}
 			d.linterDisabled = canDisable
 			if !canDisable {
-				needleLine := ln.Line() + c.Position.StartLine - 2
+				needleLine := ln.Line() + c.Position.StartLine - 1
 				d.ReportByLine(needleLine, LevelInformation, "linterError", "You are not allowed to disable linter")
+				d.ReportByLine(19, LevelInformation, "linterError", "You are not allowed to disable linter")
```
quasilyte: Probably a debug leftover?

i582: already removed

src/linttest/testdata/output-test/output_test.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+/** @linter disable */
```
quasilyte, line 3: Nice job with adding a test for all kinds of phpdoc comments formatting.
:+1:

PR #591, opened 2020-07-16
cmd: add -gitignore filter

quasilyte:
cmd: add -gitignore flag (default=false)

When -gitignore is set, NoVerify tries to respect gitignore
files and exclude ignored files from the analysis.

During the startup, we try to find a git repository root
by going up until we find a folder with .git or encounter
the FS root. We save all gitignore files we find along the way.

gitignore can't be used outside of the valid git repository.


quasilyte:
I guess we can use godirwalk PostChildrenCallback to pop gitignore files when all its children folders were traversed to avoid running a list of gitignore matchers over the entire project.

Issue #563, opened 2020-07-04
For non-void functions, forbid bare return


quasilyte:
## Code Example

```php
/**
 * @return string|null
 */
function f($x) {
  if (!$x) {
    return; // Better to write as "return null"
  }
  // ...
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Suggest explicit `return null`. Functions with void results should be ignored.

i582:
Will be tracked further in #1123.

PR #592, opened 2020-07-17
Added bare return warning in function with non-void return type

i582:
Solve: #563

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -64,10 +64,13 @@ type BlockWalker struct {
 	// state
 	statements map[node.Node]struct{}
 
-	// whether a function has a return without explit expression.
+	// whether a function has a return without explicit expression.
 	// Required to make a decision in void vs null type selection,
 	// since "return" is the same as "return null".
 	bareReturn bool
+	// stores return nodes that have no return value.
+	bareReturns []*stmt.Return
+
 	// whether a function has a return with explicit expression.
 	// When can't infer precise type, can use mixed.
 	returnsValue bool
@@ -80,6 +83,7 @@ type BlockWalker struct {
 	// shared state between all blocks
 	unusedVars   map[string][]node.Node
 	nonLocalVars map[string]struct{} // static, global and other vars that have complex control flow
+
 }
 
 func (b *BlockWalker) addStatement(n node.Node) {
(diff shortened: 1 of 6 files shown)


src/linter/report.go
```diff
@@ -457,6 +457,12 @@ function performance_test() {}`,
 			Default: true,
 			Comment: `Report linter error.`,
 		},
+
+		{
+			Name:    "bareReturnInNonVoidFunction",
```
quasilyte, line 462: Just "bareReturn" sounds better. It's generally a good practice to stay in 1-2 words per checker name.

src/linter/block.go
```diff
@@ -64,10 +64,13 @@ type BlockWalker struct {
 	// state
 	statements map[node.Node]struct{}
 
-	// whether a function has a return without explit expression.
+	// whether a function has a return without explicit expression.
 	// Required to make a decision in void vs null type selection,
 	// since "return" is the same as "return null".
 	bareReturn bool
+	// stores return nodes that have no return value.
+	bareReturns []*stmt.Return
```
quasilyte, line 72: It's more reliable to check whether a function return type `!typ.Is("void")` than depend on the `bareReturn` flag in my opinion. This will also make it possible to generate the report at the point where you're handling that return (so the slice is unnecessary in that case).

One of the corner case is this:
```php
/**
 * @return null
 */
function f() {
  return;
}
```
The `returnsValue` would be false, so you can't warn on `return` in this case even though the function is intended to be a non-void function (it returns null).

P.S. if we infer `f` function return type as void, that's a bug (it should be `null`).

src/linter/block.go
```diff
@@ -80,6 +83,7 @@ type BlockWalker struct {
 	// shared state between all blocks
 	unusedVars   map[string][]node.Node
 	nonLocalVars map[string]struct{} // static, global and other vars that have complex control flow
+
```
quasilyte, line 86: This empty line looks excessive. gofmt can't remove it, but it shouldn't be there.

src/linter/root.go
```diff
@@ -666,6 +666,11 @@ func (d *RootWalker) handleFuncStmts(params []meta.FuncParam, uses, stmts []node
 
 	switch {
 	case b.bareReturn && b.returnsValue:
+
```
quasilyte, line 669: Please avoid empty lines like these. They can separate logical blocks, but this one is superfluous.

quasilyte:
Is this PR abandoned in favor of #605?

i582:
> Is this PR abandoned in favor of #605?

Yep

Issue #594, opened 2020-07-17
Improve storage of the value of constants


i582:
A string is currently being used for storage. But this method requires parsing the string into the desired type and back, which is expensive.

PR #595, opened 2020-07-17
Improved storage of the value of constants

i582:
The way of storing values for constants has been changed. Now interface{} is used for this, which avoids unnecessary conversions from a string to a certain type and back.

Since the data is not stored in a string, methods for decoding and encoding and tests for them were written for caching.

Solve #594

### DIFF
--- a/src/meta/encoding_test.go
+++ b/src/meta/encoding_test.go
@@ -94,3 +94,59 @@ func TestTypeEncoding(t *testing.T) {
 		}
 	}
 }
+
+func TestConstantValueDecodeEncode(t *testing.T) {
+	// encode test
+	v := ConstantValue{
+		Type:  Integer,
+		Value: int64(54),
+	}
+	b, err := v.GobEncode()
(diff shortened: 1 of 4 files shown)


src/meta/encoding_test.go
```diff
@@ -94,3 +94,59 @@ func TestTypeEncoding(t *testing.T) {
...
+	v := ConstantValue{
+		Type:  Integer,
+		Value: int64(54),
+	}
+	b, err := v.GobEncode()
+	if err != nil {
+		t.Errorf("unexpected error \"%s\"", err)
+	}
+
+	if b[0] != byte(Integer) || b[1] != 53 || b[2] != 52 {
```
quasilyte, line 109: It's OK to test just encode+decode and compare that decoded value is the same as the original value.
This test with `b[1] != 53` is quite hard to comprehend and it depends on the fact that we're storing numerical values as text (hence we have 2 ascii bytes for 54 - 53 and 52).

PR #602, opened 2020-07-21
Infer types in list() over tuple

i582:
Added type inference for each variable when destructuring a variable with the `\tuple` type or a function that returns `\tuple`.
Added test.

For example:
```php
class Boo {}

/**
 * @return \tuple(int, \Boo, int, string)
 */
function foo() {
    return [5, new Boo(), 10, "gas"];
}

$tuple = foo();

[$i, $bo, $j, $str] = $tuple;

exprtype($i, "int");
exprtype($bo, "\Boo");
exprtype($j, "int");
exprtype($str, "string");
```


src/linter/block.go
```diff
@@ -2313,7 +2313,7 @@ func (b *BlockWalker) handleAssign(a *assign.Assign) bool {
 			return true
 		}
 
-		tp := solver.ExprType(b.ctx.sc, b.r.ctx.st, a.Expression)
+		tp := solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, a.Expression)
```
i582, line 2316: but there is no race, ahaha

src/linter/block.go
```diff
@@ -2302,7 +2309,19 @@ func (b *BlockWalker) handleAssign(a *assign.Assign) bool {
...
 	case *expr.List:
-		b.handleAssignList(v.Items)
+		if !meta.IsIndexingComplete() {
+			return true
+		}
+
+		tp := solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, a.Expression)
+		var findType string
+		tp.Iterate(func(t string) {
+			if strings.HasPrefix(t, "\\shape") {
```
quasilyte, line 2319: You can use raw string literal to avoid `\` escaping:
```
`\shape`
```

src/linter/block.go
```diff
@@ -2302,7 +2309,19 @@ func (b *BlockWalker) handleAssign(a *assign.Assign) bool {
 		b.checkVoidType(a.Expression)
 		b.replaceVar(v, solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, a.Expression), "assign", meta.VarAlwaysDefined)
 	case *expr.List:
-		b.handleAssignList(v.Items)
+		if !meta.IsIndexingComplete() {
+			return true
+		}
+
+		tp := solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, a.Expression)
+		var findType string
```
quasilyte, line 2317: `findType` is a verb. It's not a good variable name here in my opinion.
You can call it `shapeType` given that it's only assigned to shape types.

src/linter/block.go
```diff
@@ -2302,7 +2309,19 @@ func (b *BlockWalker) handleAssign(a *assign.Assign) bool {
...
+		tp.Iterate(func(t string) {
+			if strings.HasPrefix(t, "\\shape") {
+				findType = t
+			}
+		})
+		class, _ := meta.Info.GetClass(findType)
```
quasilyte, line 2323: If findType (or shapeType) is an empty string, there is no need to do a lookup.

src/linter/block.go
```diff
@@ -2196,9 +2196,16 @@ func (b *BlockWalker) handleAssignReference(a *assign.Reference) bool {
...
+func (b *BlockWalker) handleAssignList(items []*expr.ArrayItem, info meta.ClassInfo) {
+	for i, item := range items {
+		prop, ok := info.Properties[fmt.Sprint(i)]
+		var tp meta.TypesMap
+		if !ok {
+			tp = meta.NewTypesMap("unknown_from_list")
+		} else {
+			tp = prop.Typ
+		}
+		b.handleVariableNode(item.Val, tp, "assign")
```
quasilyte, line 2208: This method will handle all types like shape types. For non-shape types it's easier to keep the previous loop and avoid doing `fmt.Sprint` and map lookup per every item.

i582:
closed due #602

quasilyte:
Race condition is after racing you. :sweat_smile:

Issue #572, opened 2020-07-13
Golden tests fail on Windows 10


i582:
**Platform:**
Windows 10

**Description:**
Golden tests fail when running on a local machine with Windows 10.

## Actual Behavior
Tests failed

## Expected Behavior
Tests pass

## Possible reason

The following is part of the output.
Can see in it that the difference between the lines lies in the direction of slashes and the presence of `\r` in the resulting value.

```
TestGolden/embeddedrules: golden_test.go:203: results mismatch (+ have) (- want):   []string{
          	strings.Join({
          		"MAYBE   ternarySimplify: could rewrite as `(bool)$x` at testdata",
        - 		"/embeddedrules/file.php:6\r",
        + 		`\embeddedrules\file.php:6`,
          	}, ""),
          	"    sink($x ? true : false);\r",
        - 	"         ^^^^^^^^^^^^^^^^^\r",
        + 	"         ^^^^^^^^^^^^^^^^^",
          	strings.Join({
          		"MAYBE   ternarySimplify: could replace the ternary with just $x ",
          		"> $y at testdata",
        - 		"/embeddedrules/file.php:9\r",
        + 		`\embeddedrules\file.php:9`,
          	}, ""),
          	"    sink($x > $y ? true : false);\r",
        - 	"         ^^^^^^^^^^^^^^^^^^^^^^\r",
        + 	"         ^^^^^^^^^^^^^^^^^^^^^^",
          	strings.Join({
          		"MAYBE   ternarySimplify: could rewrite as `$x ?: $y` at testdata",
        - 		"/embeddedrules/file.php:12\r",
        + 		`\embeddedrules\file.php:12`,
          	}, ""),
          	"    sink($x ? $x : $y);\r",
        - 	"         ^^^^^^^^^^^^\r",
        + 	"         ^^^^^^^^^^^^",
```

PR #604, opened 2020-07-21
linter: fixed bug when GoldenTest failed on Windows

i582:
Updates #572

### DIFF
--- a/src/linttest/golden_test.go
+++ b/src/linttest/golden_test.go
@@ -275,7 +275,7 @@ func runGoldenTestsE2E(t *testing.T, targets []*goldenTest) {
 			}
 			args = append(args, target.srcDir)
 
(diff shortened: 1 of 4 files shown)


src/linttest/golden_test.go
```diff
@@ -225,9 +226,16 @@ func runGoldenTestsE2E(t *testing.T, targets []*goldenTest) {
 		return
 	}
 
+	var linterName string
+	if runtime.GOOS == "windows" {
+		linterName = "phplinter.exe"
+	} else {
+		linterName = "phplinter"
+	}
```
quasilyte, line 234: I believe Linux doesn't care about binary file extension, so maybe we can always create a file with `.exe` extension?

quasilyte: We may also want to add `phplinter.exe` to the gitignore file, just to be safe.

PR #605, opened 2020-07-22
linter: added checks for functions

### LINKED ISSUE #561: Find mismatching @return annotations
## Code Example

```php
    /**
     * Set the current local namespace.
     * @param array|null $namespace (default: null)
     * @return array|null
     */
    public function setNamespace(array $namespace = null)
    {
        $this->namespace = $namespace;
    }
```

## Actual Behavior

No warnings.

## Expected Behavior

Warning about not returning anything from a function that promised to return something (except null) in its phpdoc.

### LINKED ISSUE #563: For non-void functions, forbid bare return

### LINKED ISSUE #597: Store FQN in CurrentFunction?
Right now `ClassParseState.CurrentFunction` stores only a partial (base) function name instead of FQN.

It's probably more useful to store FQN there.

Since we store method names there as well, it might be a good time to move that to the `CurrentMethod` field? We can't store method FQN in a way that is useful for us.

This needs some research on how CurrentFunction is used. It should be trivial to get the base name out of the FQN when it's needed.

i582:
1. Added check for bare returns for functions with non-void return types. #563
2. Added check for the absence of return statement in function with `@return` annotation, which is not equal to void. #561

For this, the storage of `CurrentFunction` has been improved.
`CurrentFunction` now stores the fully qualified name of the function with its class and namespace. A separate `CurrentMethod` field was created to store the names of the methods. #597 

Fixes #561 
Fixes #563 
Fixes #597

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -528,6 +528,45 @@ func (b *BlockWalker) handleReturn(ret *stmt.Return) {
 	if ret.Expr == nil {
 		// Return without explicit return value.
 		b.bareReturn = true
+
+		if !meta.IsIndexingComplete() {
+			return
+		}
+
+		funcName := b.r.ctx.st.CurrentFunction
+		currentClass := b.r.ctx.st.CurrentClass
+		currentMethod := b.r.ctx.st.CurrentMethod
+
+		if len(currentClass) != 0 {
+			funcName = currentMethod
+
+			class, ok := meta.Info.GetClass(currentClass)
+			if !ok {
+				return
+			}
+			fun, ok := class.Methods.Get(funcName)
+			if !ok {
+				return
+			}
+
+			if !fun.Typ.Is("void") {
+				b.r.Report(ret, LevelWarning, "bareReturn", "Replace 'return' with 'return null'")
+			}
+		} else {
+			fun, ok := meta.Info.GetFunction(funcName)
+			if !ok {
+				return
+			}
+
+			if !fun.Typ.Is("void") {
+				b.r.Report(ret, LevelWarning, "bareReturn", "Replace 'return' with 'return null'")
+			}
+		}
+
+		// for _, el := range meta.Info.allFunctions.H {
+		// 	fmt.Println(el.Name)
+		// }
+
 		return
 	}
 	b.returnsValue = true
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -64,10 +64,14 @@ type BlockWalker struct {
 	// state
 	statements map[node.Node]struct{}
 
-	// whether a function has a return without explit expression.
+	// whether a function has a return without explicit expression.
 	// Required to make a decision in void vs null type selection,
 	// since "return" is the same as "return null".
 	bareReturn bool
+	// when the function contains at least one return or yield.
+	containsReturnOrYield bool
+	// when the function contains throw.
+	containsThrow bool
 	// whether a function has a return with explicit expression.
 	// When can't infer precise type, can use mixed.
 	returnsValue bool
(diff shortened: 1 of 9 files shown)


src/linter/block.go
```diff
@@ -64,10 +64,14 @@ type BlockWalker struct {
 	// state
 	statements map[node.Node]struct{}
 
-	// whether a function has a return without explit expression.
+	// whether a function has a return without explicit expression.
 	// Required to make a decision in void vs null type selection,
 	// since "return" is the same as "return null".
 	bareReturn bool
+	// when the function contains at least one return or yield.
+	containsReturnOrYield bool
```
YuriyNasretdinov, line 72: exitFlags should already contain this information in the way you want?

src/linter/block.go
```diff
@@ -525,9 +544,50 @@ func (b *BlockWalker) handleFunction(fun *stmt.Function) bool {
...
+			if !ok {
+				return
+			}
+
+			if !fun.Typ.Is("void") {
+				b.r.Report(ret, LevelWarning, "bareReturn", "Replace 'return' with 'return null'")
+			}
+		}
+
+		// for _, el := range meta.Info.allFunctions.H {
```
YuriyNasretdinov, line 587: Probably remove this?

src/linter/root.go
```diff
@@ -698,7 +698,12 @@ func (d *RootWalker) handleFuncStmts(params []meta.FuncParam, uses, stmts []node
 		prematureExitFlags = cleanFlags
 	}
 
+	// if the function contains only one statement and this statement is throw.
+	oneThrowStatementFunction := b.containsThrow && len(stmts) == 1
```
YuriyNasretdinov, line 702: Why not two? Ten? Why this special case only?

src/linttest/testdata/mustache/golden.txt
```diff
@@ -52,6 +52,21 @@ MAYBE   phpdoc: Missing PHPDoc for "getHelperName" public method at testdata/mus
 MAYBE   phpdoc: Missing PHPDoc for "getTemplateName" public method at testdata/mustache/src/Mustache/Exception/UnknownTemplateException.php:34
     public function getTemplateName()
                     ^^^^^^^^^^^^^^^
+WARNING bareReturn: Replace 'return' with 'return null' at testdata/mustache/src/Mustache/Parser.php:207
```
YuriyNasretdinov, line 55: return; and return null; already mean the same thing. Maybe provide more context here on why exactly do you ask to replace it in this function?

src/linter/root.go
i582: in `basic_test.go`:

```
// Functions of this kind are very common in Symfony, so it makes sense not to consider this 
// a loss of return, although in fact the @return annotation is not entirely correct.
```

src/linter/block.go
i582: I need to know if there is a `return` or `yield` in the body of the function in order to accurately determine `"mismatchingReturn"`.  I'm not sure that exitFlags holds the information I need. Comment mentions other things `// if function has exit / die / throw, then ExitFlags will be <> 0`.

src/linttest/testdata/mustache/golden.txt
i582: Probably we can provide additional information after the first sentence?
For example:
```
<critical> WARNING bareReturn: Replace 'return' with 'return null'. The '\f' function has @return annotation with non-'void' value, so 'return null' is preferred. at C:/projects/noverify/example/index.php:7
    return;
    ^^^^^^^
```

i582:
Closed until better times.

PR #606, opened 2020-07-23
linter: improved "deprecated" error message

### LINKED ISSUE #542: If @removed tag presents, use that info to make deprecated message more precise
## Code Example

```php
    /**
     * Gets the class type hinted for the parameter as a ReflectionClass object.
     *
     * @link https://php.net/manual/en/reflectionparameter.getclass.php
     * @return ReflectionClass|null A {@see ReflectionClass} object.
     * @deprecated 7.4
     * @removed 8.0
     */
    public function getClass()
    {
    }
```

## Actual Behavior

```
deprecated: Call to deprecated method {\ReflectionParameter}->getClass() (7.4)
```

## Expected Behavior

```
deprecated: Call to deprecated method {\ReflectionParameter}->getClass() (7.4, removed in 8.0)
```

Low-priority.

i582:
Now, if there is an `@removed` annotation, then it will be included in the message as additional information. #542 
Hash version increased.
Added new tests.

Fixes #542


quasilyte:
I would change tests to have more diverse tags.
In the wild, there can be very unstructured tags:
```
@removed
@removed a long time ago
@removed somewhere around 4.5
@removed yii2.5

@deprecated
@deprecated use Foo instead
@deprecated in 4.5

etc, etc (you can check out projects in open source to get some inspirations)
```
You can't predict which format is going to be used, but you must be sure that different formats don't break the desired behavior.

src/meta/metainfo.go
```diff
@@ -299,6 +299,62 @@ type FuncParam struct {
...
+}
+
+func (p PhpDocInfo) StringForMethod(exprType TypesMap, methodName string) string {
+	if methodName == "" {
+		return ""
+	}
+	return p.string("", exprType, methodName)
+}
+
+func (p PhpDocInfo) string(funcName string, exprType TypesMap, methodName string) string {
```
quasilyte, line 321: I'm not that comfortable with a method name that is the same as builtin type name.
But please don't call it `String()` as that method usually more general than giving a specific printing format for reports.
You may format the message somewhere in the linter code.

i582:
Closed until better times.

PR #607, opened 2020-07-23
linter: added reports for static and non-public magic methods.

### LINKED ISSUE #551: Give a warning for non-public __call (and other)
## Code Example

```php
<?php

class A {
  private function __call($method, $args) {
    var_dump([$method => $args]);
  }
}

class B extends A {
}

$b = new B();
$b->dd(1);
```

## Actual Behavior

No warnings.

If we run that code, PHP will give a warning:
```
The magic method __call() must have public visibility and cannot be static
```

## Expected Behavior

Report that __call should be a public non-static method.

i582:
Added reports for static and non-public magic methods.
Added test.

Fixes #551


src/linter/root.go
```diff
@@ -783,6 +783,39 @@ func (d *RootWalker) parseMethodModifiers(meth *stmt.ClassMethod) (res methodMod
 	return res
 }
 
+func (d *RootWalker) checkMagicMethod(meth node.Node, name string, modif methodModifiers) {
+	switch name {
+	case " __construct",
```
YuriyNasretdinov, line 788: Constructors can be private and that's not an error

src/linter/report.go
```diff
@@ -455,6 +455,20 @@ function performance_test() {}`,
...
+			Name:    "magicMethod",
+			Default: true,
+			Comment: `Report error in magic methods.`,
+			Before: `class Foo {
+  private function __call($method, $args) {} // The magic method __call() must have public visibility
+  static function __set($name, $value) {} // The magic method __set() cannot be static
+}`,
+			After: `class Foo {
+  public function __call($method, $args) {} // Ok
+  function __set($name, $value) {} // Ok
```
quasilyte, line 469: It's generally a bad practice to omit access modifiers in classes. "public" is implied, but our examples shouldn't display a code that smells. :)

src/linter/report.go
```diff
@@ -471,6 +471,20 @@ function performance_test() {}`,
 			Default: true,
 			Comment: `Report linter error.`,
 		},
+
+		{
+			Name:    "magicMethod",
```
quasilyte, line 476: I would suggest calling it `magicMethodDecl` as it doesn't check all kinds of mistakes, but only declaration-related ones (visibility, instance/static modifiers).

src/linter/report.go
```diff
@@ -471,6 +471,20 @@ function performance_test() {}`,
...
+			Default: true,
+			Comment: `Report error in magic methods.`,
```
quasilyte, line 478: "Report issues in magic method declarations."

src/linter/root.go
```diff
@@ -784,6 +784,53 @@ func (d *RootWalker) parseMethodModifiers(meth *stmt.ClassMethod) (res methodMod
...
+			d.Report(meth, LevelError, "magicMethod", "The magic method %s() cannot be static", name)
+		}
+
+	case "__callStatic":
+		// must be static and public
+		if !modif.static {
+			d.Report(meth, LevelError, "magicMethod", "The magic method %s() must be static", name)
+		}
+		if modif.accessLevel != meta.Public {
+			d.Report(meth, LevelError, "magicMethod", "The magic method %s() must have public visibility", name)
```
quasilyte, line 818: Now I see that we haven't updated checker name in Report calls.

PR #608, opened 2020-07-23
linter: improved the message for 'unused' in foreach

### LINKED ISSUE #256: Better warning for "unused foreach key variable"
## Code Example

```php
foreach ($x as $i => $v) {
  var_dump($v);
}
```

## Actual Behavior

Warning: "Unused variable i (use $_ to ignore this inspection)".

## Expected Behavior

While actual behavior is correct, the message itself is a little confusing.
Instead, it would be nice to have:

> foreach key $i is unused, can simplify `$i => $v`  to just `$v`.

If code was fixed to use `$_` instead, we might want to introduce a separate check that suggests simplifying it to the value-only foreach loop.

i582:
Improved the message for `'unused'` in `foreach`, now instead of suggesting a replacement for `$_`, a simplification is suggested.

Fixes #256


src/linter/block.go
```diff
@@ -1558,6 +1558,26 @@ func (b *BlockWalker) handleForeach(s *stmt.Foreach) bool {
...
+	_, ok = b.unusedVars[key.Name]
+	if ok {
+		variable, ok := s.Variable.(*node.SimpleVar)
+		if !ok {
+			return false
+		}
+
+		delete(b.unusedVars, key.Name)
+
+		b.r.Report(s.Key, LevelError, "foreachSimplify", "foreach key $%s is unused, can simplify $%s => $%s to just $%s", key.Name, key.Name, variable.Name, variable.Name)
```
quasilyte, line 1578: It looks like it shouldn't be a new diagnostic as it checks the same thing as before. The only difference is that it tries to give more human-readable (and more actionable) warning message.

@YuriyNasretdinov what do you think about reporting this as "unused"?

YuriyNasretdinov: Yeah, I guess unused would be appropriate here.

quasilyte:
This is strange that diff includes something that was already merged into the master.

PR #613, opened 2020-07-28
linttest: added default stubs for all tests

### LINKED ISSUE #531: Improve builtin const special handling in undefined check
Right now it tries to checks if true/false was written in anything other than lowercase (e.g. TRUE).

The problem arises in tests where we do not preload these constants, so we get a weird warning:
```
ERROR undefined: Use true instead of true
```
We should either preload builtin consts for all tests or give a better warning for this particular case.

i582:
These stubs add constants and some functions that can now be used in tests without writing their prototypes in the test itself.

Fixes #531


i582:
aaaaaa, it happened again

src/cmd/stubs/phpstorm_stubs.go
```diff
@@ -7531,10 +7531,10 @@ func AssetInfo(name string) (os.FileInfo, error) {
 }
 
 // AssetNames returns the names of the assets.
-func AssetNames() []string {
-	names := make([]string, 0, len(_bindata))
+func AssetNames() map[string]struct{} {
```
quasilyte, line 7534: This is an auto-generated code. Please don't edit it.

PR #614, opened 2020-07-28
Infer types in list() over tuple

i582:
Added type inference for each variable when destructuring a variable with the `\tuple` type or a function that returns `\tuple`.
Added test.

For example:
```php
class Boo {}

/**
 * @return \tuple(int, \Boo, int, string)
 */
function foo() {
    return [5, new Boo(), 10, "gas"];
}

$tuple = foo();

[$i, $bo, $j, $str] = $tuple;

exprtype($i, "int");
exprtype($bo, "\Boo");
exprtype($j, "int");
exprtype($str, "string");
```

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -2322,7 +2322,7 @@ func (b *BlockWalker) handleAssign(a *assign.Assign) bool {
 		tp := solver.ExprType(b.ctx.sc, b.r.ctx.st, a.Expression)
 		var shapeType string
 		tp.Iterate(func(t string) {
-			if strings.HasPrefix(t, `\shape`) {
+			if strings.HasPrefix(t, `\shape$`) {
 				shapeType = t
 			}
 		})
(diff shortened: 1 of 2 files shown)


src/linter/block.go
```diff
@@ -2302,7 +2315,27 @@ func (b *BlockWalker) handleAssign(a *assign.Assign) bool {
...
 	case *expr.List:
-		b.handleAssignList(v.Items)
+		if !meta.IsIndexingComplete() {
+			return true
+		}
+
+		tp := solver.ExprType(b.ctx.sc, b.r.ctx.st, a.Expression)
+		var shapeType string
+		tp.Iterate(func(t string) {
+			if strings.HasPrefix(t, `\shape`) {
```
quasilyte, line 2325: Not related to a data race, but it should be `\shape$` to avoid collisions with `\shape` class (if user defined it). Auto-generated shape types have `$` in their name.

PR #619, opened 2020-07-30
linter: collect and expose AST node path

quasilyte:
For every visited node, push it to the node path.
Pop node when we leave the node.

If EnterNode() returns false, we need to do a pop
since LeaveNode() will not be called in that case.

Where we can use it?
- Get current node parent.
- Find whether this node is an argument of specific func call.
- Find whether this node located under if/for/etc.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -58,6 +58,8 @@ type BlockWalker struct {
 
 	custom []BlockChecker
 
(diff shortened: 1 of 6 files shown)


quasilyte:
I don't think I covered all branch nodes, so there probably will be 1 more update to make it more exhaustive.

Issue #609, opened 2020-07-23
Type inference for closures


i582:
Currently, the type for the closure arguments is not inferred, but that would be helpful.

There are two main direction, I think:
1. handle only built-in functions like sorting.
2. try to handle even custom functions.

## Code Example

```php
$d = [10, 56, 32, 12];

usort($d, function($a, $b)
{
  exprtype($a, "int");
  exprtype($b, "int");

  if ($a == $b) {
    return 0;
  }
  return ($a < $b) ? -1 : 1;
});
```

## Actual Behavior
The types are wrong.

## Expected Behavior
The types are correct.

quasilyte:
Custom functions behavior can't be tracked without good flow analysis which we currently don't have.

I would start from the builtin functions like `array_map`.

Param type hints help in some cases, but PHP doesn't have typed arrays support, so one can't express `[]Foo` type, for example. This is why we need this feature.

In normal functions, phpdoc can help, but it's less obvious how to associate free-floating comments with the lambda.

PR #620, opened 2020-07-31
linter: added the ability to type inference for arguments to callback function when calling a function with this callback.

i582:
Each required function must be registered in the `ClosureCallerInfo.Model()` function

> Supported only if callback is written in the arguments of the called function!

```php
class Foo { public function f() {} }
class Boo { public function b() {} }

/** @return Foo[] */
function return_foo() { return []; }
/** @return Boo[] */
function return_boo() { return []; }

$foo_array = return_foo();
$boo_array = return_boo();

usort($foo_array, function($a, $b) {
  $a->f();
  $b->f();
  exprtype($a, "\Foo");
  exprtype($b, "\Foo");
});

array_map(function($a) {
  $a->f();
  exprtype($a, "\Foo");
}, $foo_array);

array_map(function($a, $b) {
  $a->f();
  $b->b();
  exprtype($a, "\Foo");
  exprtype($b, "\Boo");
}, $foo_array, $boo_array);

array_walk($foo_array, function($a, $b) {
  $a->f();
  exprtype($a, "\Foo");
  exprtype($b, "mixed");
});

some_function_without_model(function($b) {
  $b->f();
  exprtype($b, "mixed");
}, $d);


// Not supported
function callback($a) {
  exprtype($a, "mixed");
}

// Not supported
$callback = function($a) {
  exprtype($a, "mixed");
};

array_map($callback, $d);
array_map('callback', $d);
```

## Adding a new function.

The new function is added in the `solver.go` file in the `Model` (line 636) function.

To add a new function, add a new case with the name of the required function.
Write `return ci.model (...)`, where instead of ellipsis, write the parameters you want.

### Let's say we add the `uasort` function.
The types in the callback depend on the type of the first argument, so we write `BaseTypeIndex: 0` in the parameters, thereby setting the required type for the arguments.
The callback function contains two arguments, the type of which depends on the first argument of the function, so we write `CountTypedArgs: 2`, thereby saying that the callback will have two arguments with the types of the first argument.

Thus, we get the following:

```go
return ci.model(ArgsModel {BaseTypeIndex: 0, CountTypedArgs: 2})
```

### However, suppose we want the `array_walk` function. 
In it, the first argument to callback also depends on the first argument to the function, but the second argument is a key whose type we cannot infer. We need to make it mixed. 
To fill the remaining types with mixed types, add `CountAllArgs: 2`, which says that the number of arguments is 2, and since` CountTypedArgs: 1`, the remaining argument will be mixed.

```go
return ci.model(ArgsModel{BaseTypeIndex: 0, CountTypedArgs: 1, CountAllArgs: 2})
```

### What about functions like `array_map`?

The `array_map` function can accept several arrays after the callback, while in the callback the types of the arguments correspond to the types of the passed arrays.

```php
array_map(function($a, $b) {
  $a->f();
  $b->b();
  exprtype($a, "\Foo");
  exprtype($b, "\Boo");
}, $foo_array, $boo_array);
```

Therefore, we must be able to specify this behavior. 
For this, the `BaseTypeIndexShiftCount` field is used. It means how many arguments to the callback depends on the passed arguments to the function.

If we talk about `array_map`, then this can be calculated by subtracting one from the number of function arguments, thereby obtaining the number of required shifts.

```go
count := len(ci.FunctionArgs) - 1
return ci.model(ArgsModel{
  BaseTypeIndex:           1,
  CountTypedArgs:          1,
  BaseTypeIndexShiftCount: count,
})
```

#609 (1)


i582:
Added the ability to infer types for arguments to functions such as `array_map`, where the type of the n callback argument depends on the type of the n + 1 argument of the calling function.

Example:
```php
class Foo {
   public function f() {}
}

class Boo {
  public function b() {}
}

class Poo {
  public function p() {}
}

$arrFoo = [new Foo(), new Foo()];
$arrBoo = [new Boo(), new Boo()];
$arrPoo = [new Poo(), new Poo()];

array_map(function($a)
{
	$a->f();
	exprtype($a, "\Foo");
}, $arrFoo);

array_map(function($a, $b)
{
	$a->f();
	$b->b();
	exprtype($a, "\Foo");
	exprtype($b, "\Boo");
}, $arrFoo, $arrBoo);

array_map(function($a, $b, $c)
{
	$a->f();
	$b->b();
	$c->p();
	exprtype($a, "\Foo");
	exprtype($b, "\Boo");
	exprtype($c, "\Poo");
}, $arrFoo, $arrBoo, $arrPoo);
```

src/meta/scope.go
```diff
@@ -40,6 +40,9 @@ type Scope struct {
 	vars             map[string]*scopeVar // variables declared in the scope
 	inInstanceMethod bool
 	inClosure        bool
+
+	CallerFunction     FuncInfo    // function that is called with this closure
+	CallerFunctionArgs []node.Node // and the arguments with which it is called
```
quasilyte, line 45: Can we actually store `[]node.Node` inside the cache?

src/linter/cache.go
```diff
@@ -38,7 +38,8 @@ import (
 //     38 - replaced TypesMap.immutable:bool with flags:uint8.
 //          added mapPrecise flag to mark precise type maps.
 //     39 - added new field Value in ConstantInfo
-const cacheVersion = 39
+//     40 - added array_map call in test
+const cacheVersion = 40
```
quasilyte, line 42: You did not change the cache layout. The version can be left alone.

src/linter/cache_test.go
```diff
@@ -86,6 +86,15 @@ function as_shape($x) {
   return $x;
 }
 
+class Foo {}
+
+function infer_types_in_callback() {
+  /** @var Foo[] $a */
+  $a = [];
+
+  array_map(function($item) {}, $a);
```
quasilyte, line 95: I would like the array_map result to be used. Tests should have valid code that doesn't trigger any warnings (preferably).

src/meta/typesmap.go
```diff
@@ -376,3 +376,31 @@ func (m TypesMap) Iterate(cb func(typ string)) {
...
+		if !strings.HasSuffix(typ, "[]") {
+			return false
+		}
+	}
+	return true
+}
+
+// Returns the base type for an array. T[] -> T, T[][] -> T[].
+// Not for *Lazy* type.
+func (m TypesMap) ArrayBaseType() (TypesMap, bool) {
```
quasilyte, line 393: Please use "ElemType" wording here instead of "BaseType".

quasilyte:
@YuriyNasretdinov could you take a look as well? :)

src/linter/block.go
```diff
@@ -1193,7 +1198,19 @@ func (b *BlockWalker) enterClosure(fun *ir.ClosureExpr, haveThis bool, thisType
 		delete(b.unusedVars, v.Name)
 	}
 
-	params, _ := b.r.parseFuncArgs(fun.Params, phpDocParamTypes, sc)
+	// find the types for the arguments of the function that contains this closure
+	var funcArgs []meta.TypesMap
+	for _, arg := range b.ctx.callerFunctionArgs {
+		tp := solver.ExprType(parentScope, b.r.ctx.st, arg.(*ir.Argument).Expr)
```
YuriyNasretdinov, line 1204: You cannot call ExprType() during the indexing phase

src/meta/typesmap.go
```diff
@@ -376,3 +376,31 @@ func (m TypesMap) Iterate(cb func(typ string)) {
 		cb(k)
 	}
 }
+
+// Checks that the map contains only arrays.
+// Not for *Lazy* type.
+func (m TypesMap) ContainsOnlyArrays() bool {
```
YuriyNasretdinov, line 382: You can probably just ignore the types that are not arrays. In PHP it can be really hard to know the types for sure anyway so it's also hard to guarantee that the type will only be an array and nothing else.

src/linter/block.go
YuriyNasretdinov: Basically this invalidates the whole implementation. You cannot use solved types during this stage so you can't check that the type of the arguments are only arrays and you don't need to as well.

i582:
Closed due to #690

PR #625, opened 2020-08-02
linter: added type inference for arrays of complex types

i582:
```php
class Foo {}
$a = [new Foo(), new Foo()];
// exprtype($a, \Foo[]);
$b = [[new Foo(), new Foo()],[new Foo(), new Foo()]];
// exprtype($b, "\Foo[][]");
```


quasilyte:
@YuriyNasretdinov as far as I understand, it doesn't continue to collect types if element types do mismatch.
So it always returns a map of 1 element (or false + nil map if it can't find that element type).

PR #632, opened 2020-08-05
linter: added arrow functions processing in linter

i582:
Added support for arrow functions at the linter level.


quasilyte:
@YuriyNasretdinov any final notes?
Storing []BlockWalker doesn't look elegant, but at least we don't copy scopes eagerly.
Arrow functions have different scoping/enclosing rules than closures, so they required some kludges to work correctly.
If you're OK with it, we might merge it today.

quasilyte:
I don't have other change suggestions right now. Let's resolve conflicts and merge this.

PR #638, opened 2020-08-06
linter: added embedded rules for assign

i582:
Example:
```php
$a = $a + $b => $a += $b
$a = $a - $b => $a -= $b
$a = $a * $b => $a *= $b
$a = $a / $b => $a /= $b
$a = $a % $b => $a %= $b

$a = $a . $b => $a .= $b

$a = $a & $b => $a &= $b
$a = $a | $b => $a |= $b
$a = $a ^ $b => $a ^= $b
$a = $a << $b => $a <<= $b
$a = $a >> $b => $a >>= $b
$a = $a ?? $b => $a ??= $b
```

For Coalesce added case to `phpgrep` to handle `assign.Coalesce`.

Added tests.


src/cmd/embeddedrules/rules.php
```diff
@@ -31,3 +31,11 @@ function ternarySimplify() {
    */
   isset($x) ? $x : $y;
 }
+
+function coalesceSimplify() {
+  /**
+   * @maybe could rewrite as `$x ??= $y`
+   * @fix $x ??= $y
+   */
+  $x = $x ?? $y;
```
quasilyte, line 40: I believe `$x` needs to be pure in this rule.

src/cmd/embeddedrules/rules.php
```diff
@@ -31,3 +31,12 @@ function ternarySimplify() {
    */
   isset($x) ? $x : $y;
 }
+
+function coalesceSimplify() {
```
quasilyte, line 35: What about calling it `assignOp` and add things like:
```
$x = $x + $y => $x += $y
...etc
```

PR #639, opened 2020-08-06
linter: added handling of type hints for class property

i582:


src/tests/exprtype/exprtype_test.go
```diff
@@ -1868,6 +1868,56 @@ exprtype($tuple_int_str, "unknown_from_list");
...
+
+class Foo {
+  public static int $int;
+  public static float $float;
+  public static string $string;
+  public static bool $bool;
+  public static array $array;
+  public static callable $callable;
+  public static iterable $iterable;
+  public static object $object;
```
quasilyte, line 1884: I would replace at least some of these test cases with non-static typed properties tests.

quasilyte:
Please add tests for type combining when phpdoc present as well. You have append for types, so they are expected to become a union type. As a corner case, `T` + `T` should remain `T` (this is handled by maps automatically, but it's nice to have a test for it).

PR #645, opened 2020-08-07
linter: added check for the correct number of arguments for magic methods

i582:
Fixed errors in tests (`TestNonPublicMagicMethods`, `TestIssue8`) that appeared due to a new check.

### DIFF
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -779,14 +779,13 @@ func (d *RootWalker) parseMethodModifiers(meth *stmt.ClassMethod) (res methodMod
 }
 
 func (d *RootWalker) checkMagicMethod(meth node.Node, name string, modif methodModifiers, countArgs int) {
+	const Any = -1
 	var (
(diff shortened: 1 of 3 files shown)


src/linter/root.go
```diff
@@ -778,33 +778,55 @@ func (d *RootWalker) parseMethodModifiers(meth *stmt.ClassMethod) (res methodMod
...
-func (d *RootWalker) checkMagicMethod(meth node.Node, name string, modif methodModifiers) {
+func (d *RootWalker) checkMagicMethod(meth node.Node, name string, modif methodModifiers, countArgs int) {
 	var (
 		canBeStatic    bool
 		canBeNonPublic bool
 		mustBeStatic   bool
+
+		mustHaveZeroArguments bool
+		mustHaveOneArgument   bool
+		mustHaveTwoArguments  bool
```
quasilyte, line 789: What about a single `numArgsExpected int` variable?

PR #660, opened 2020-08-11
linter: infer types in list() over shape

i582:
Added type inference for each variable when destructuring variable with the `\shape` type or function that returns `\shape`.

For example:
```php
/** @return shape(key:\Foo, key2:float, key3:int) */
function asShapeWithStringKey() { return []; }

function foo() {
  ["key" => $a, "key2" => $b, "key3" => $c] = asShapeWithStringKey();

  exprtype($a, "\Foo");
  exprtype($b, "float");
  exprtype($c, "int");
}
```

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1618,11 +1618,11 @@ func (b *BlockWalker) handleAssignShapeToList(items []*ir.ArrayItemExpr, info me
 				key = keyNode.Value
 			case *ir.Dnumber:
 				key = keyNode.Value
-			default:
-				key = "unhandled_type_of_key"
 			}
 
-			prop, ok = info.Properties[key]
+			if key != "" {
+				prop, ok = info.Properties[key]
+			}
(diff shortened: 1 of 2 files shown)


src/linter/block.go
```diff
@@ -1604,18 +1604,42 @@ func (b *BlockWalker) handleAssignReference(a *ir.AssignReference) bool {
...
+			var key string
+			switch keyNode := item.Key.(type) {
+			case *ir.String:
+				key = unquote(keyNode.Value)
+			case *ir.Lnumber:
+				key = keyNode.Value
+			case *ir.Dnumber:
+				key = keyNode.Value
+			default:
+				key = "unhandled_type_of_key"
```
quasilyte, line 1622: I would keep the empty string and avoid lookup for `unhandled_type_of_key`.
It's also suspicious if there is a `unhandled_type_of_key` key in a shape. ;)

PR #663, opened 2020-08-12
linter: added processing of array unpacking (...$arr)

i582:


quasilyte:
I think it's a solver package change, not linter package change. I'll edit the commit message.

PR #689, opened 2020-08-19
linter: added type inference for list arguments in foreach for tuple and shape.

i582:
For example:

```php
/**
 * @return tuple(int, string)
 */
function asTuple() {
    return [];
}

function f() {
  // array of tuples
  $tuples = [asTuple(), asTuple(), asTuple()];
  foreach ($tuples as list($a, $b)) {
    exprtype($a, "int");
    exprtype($b, "string");
  }
}
```

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1080,8 +1080,36 @@ func (b *BlockWalker) handleForeach(s *ir.ForeachStmt) bool {
 
 			b.handleVariableNode(s.Key, meta.TypesMap{}, "foreach_key")
 			if list, ok := s.Variable.(*ir.ListExpr); ok {
-				for _, item := range list.Items {
-					b.handleVariableNode(item.Val, meta.TypesMap{}, "foreach_value")
+				if !meta.IsIndexingComplete() {
+					return
+				}
+
+				tp := solver.ExprType(b.ctx.sc, b.r.ctx.st, s.Expr)
+				var shapeType string
+				tp.Iterate(func(t string) {
+					if strings.HasPrefix(t, `\shape$`) {
+						shapeType = t
+					}
+				})
+
+				isArrayOfShape := strings.HasSuffix(shapeType, "[]")
+
+				if !isArrayOfShape {
+					for _, item := range list.Items {
+						b.handleVariableNode(item.Val, meta.MixedType, "foreach_value")
+					}
+				} else {
+					shapeType = strings.TrimSuffix(shapeType, "[]")
+
+					var class meta.ClassInfo
+					var ok bool
+					var isShape bool
+					if shapeType != "" {
+						class, ok = meta.Info.GetClass(shapeType)
+						isShape = ok
+					}
+
+					b.handleAssignList(list.Items, class, isShape)
 				}
 			} else {
 				b.handleVariableNode(s.Variable, meta.TypesMap{}, "foreach_value")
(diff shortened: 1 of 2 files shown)


src/linter/block.go
```diff
@@ -1080,8 +1080,36 @@ func (b *BlockWalker) handleForeach(s *ir.ForeachStmt) bool {
...
-					b.handleVariableNode(item.Val, meta.TypesMap{}, "foreach_value")
+				if !meta.IsIndexingComplete() {
+					return
+				}
+
+				tp := solver.ExprType(b.ctx.sc, b.r.ctx.st, s.Expr)
+				var shapeType string
+				tp.Iterate(func(t string) {
+					if strings.HasPrefix(t, `\shape$`) {
+						shapeType = t
```
YuriyNasretdinov, line 1091: There can be multiple types like this

src/linter/block.go
```diff
@@ -1080,8 +1080,36 @@ func (b *BlockWalker) handleForeach(s *ir.ForeachStmt) bool {
 
 			b.handleVariableNode(s.Key, meta.TypesMap{}, "foreach_key")
 			if list, ok := s.Variable.(*ir.ListExpr); ok {
-				for _, item := range list.Items {
-					b.handleVariableNode(item.Val, meta.TypesMap{}, "foreach_value")
+				if !meta.IsIndexingComplete() {
```
YuriyNasretdinov, line 1083: We need to solve this using lazy types. If the result of a foreach is returned in this function and is used in another one, this logic will not allow the type to propagate all the way. The following should also work:

```
<?php
class Foo { public function hello () {} }

function shaper() {
return shape(1, new Foo());
}

function foreacher() {
foreach ([shaper()] as list($_, $foo)) return $foo;
}

function fooer() {
foreacher()->hello(); // <- no errors here
}
```

YuriyNasretdinov (changes requested):

src/linter/block.go
i582: If we add the necessary comments, then this code passes the test.
```php
<?php
class Foo { public function hello () {} }

/**
 * @return shape(Key1:int, Key2:\Foo)
 */
function shaper() {
	return ["Key1" => 1, "Key2" => new Foo()];
}

/**
 * @return \Foo
 */
function foreacher() {
	foreach ([shaper()] as list("Key1" => $_, "Key2" => $foo)) {
		exprtype($foo, "\Foo");
		return $foo;
	}
}

function fooer() {
	$f = foreacher();
	exprtype($f, "\Foo");
	$f->hello(); // <- no errors here
}
```

YuriyNasretdinov: Sure, but noverify doesn't require you to add phpdoc-comments in cases when it can infer types. Requiring annotations to do such things seems impractical to me and in "real" code annotations for simple functions are often omitted so I would prefer that we try to infer as much information from the code as we can so that you don't have to write annotations everywhere just to please the linter.

i582: I agree, however, in this code, the `foreacher` function will not always return the `Foo` type, for example, if `shaper` returns an empty array, then the function will return `void`, which the linter correctly complains about if the `foreacher` does not have phpdoc.

quasilyte:
What's the status on this?
Can this be merged if conflicts are resolved?

i582:
Closed until better times.

PR #694, opened 2020-08-20
linter: fix node path for catch stmt

quasilyte:
We don't call Walk() on catch stmt, so ir.CatchStmt never
gets into the node path. This makes Conditional() never
return true for ir.CatchStmt case.


quasilyte:
I'll check whether we can call Walk on catch stmt and get rid of the ugly push/pop hack.

PR #700, opened 2020-08-21
constant: rename constant value

i582:


src/linter/cache.go
```diff
@@ -37,7 +37,7 @@ import (
 //          changed meta.scopeVar bool fields representation
 //     38 - replaced TypesMap.immutable:bool with flags:uint8.
 //          added mapPrecise flag to mark precise type maps.
-//     39 - added new field Value in ConstantInfo
+//     39 - added new field Value in ConstInfo
 //     40 - changed string const value storage (no quotes)
 //     41 - const-folding affected const definition values
 //     42 - bool-typed consts are now stored in meta info
```
quasilyte, line 43: Since cache serialization is changed (gob encodes type names as well), cache version should be bumped.
I would keep ConstaintInfo entry for 39th version and add a new version that mentions that the type was renamed (hence the new version).

quasilyte:
Looks good, but I would bump the cache version.

PR #704, opened 2020-08-24
linter: improved checking for duplicate cases

i582:
Added support for constants.


src/meta/const_value.go
```diff
@@ -203,6 +203,14 @@ func (c ConstantValue) String() string {
 	return fmt.Sprintf("%s(%v)", c.Type, c.Value)
 }
 
+func (c ConstantValue) StringValue() string {
```
quasilyte, line 206: I thought that we already had this method...

quasilyte: Ah, I see. This one does something different.

src/meta/const_value.go
```diff
@@ -203,6 +203,14 @@ func (c ConstValue) String() string {
 	return fmt.Sprintf("%s(%v)", c.Type, c.Value)
 }
 
+func (c ConstValue) StringValue() string {
```
quasilyte, line 206: If we move this method to some util function in linter package, I can approve this PR.

quasilyte (approved):

PR #718, opened 2020-08-26
tests/golden: added the ability to test quickfix in embedded rules

### LINKED ISSUE #715: Add quickfix (-fix) tests (esp. for dynamic rules)
Right now we don't have any tests that verify that our linter performs the correct fixes.
We can use `golden` tests framework (or just approach in general) to implement this kind of tests.

i582:
1. Create your own file for each rule. In it, create a function named like the rule and add the necessary tests.
2. Start testing. For each `.php` file, file with the extension `.php.fix.expected` will be created where the results of fixing the `.php` file will be automatically placed. Thus, we do not have to do it ourselves.
3. Check that everything is fixed correctly in the `.php.fix.expected` file.

Fixes #715


quasilyte:
> tests/golden: added the ability to test embedded rules

I would say that this is related to the quickfix testing, not the embedded rules testing (which we already have).

src/tests/golden/testdata/quickfix/binaryOp.php
```diff
@@ -0,0 +1,16 @@
...
+    $x = $x / $y;
+    $x = $x % $y;
+    $x = $x & $y;
+    $x = $x | $y;
+    $x = $x ^ $y;
+    $x = $x << $y;
+    $x = $x >> $y;
+    $x = $x . $y;
+    $x = $x ?? $y;
+}
```
quasilyte, line 16: It's a good practice to have a trailing newline inside a file.

src/linttest/linttest.go
```diff
@@ -213,6 +223,8 @@ func (s *Suite) RunLinter() []*linter.Report {
 		}
 	}
 
+	linter.ApplyQuickFixes = s.ApplyQuickFixes
```
quasilyte, line 226: In the current architecture, you don't need a suite field to set the global variable.
I would set the `linter.ApplyQuickFixes` inside the quickfixes test.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,126 @@
...
+	s := NewQuickFixTest(t, `testdata/quickfix`)
+	s.runQuickFixTest()
+}
+
+const (
+	expectedExtension = ".fix.expected"
+	fixExtension      = ".fix"
+)
+
+type QuickFixTest struct {
```
quasilyte, line 25: This type doesn't need to be exported.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,126 @@
...
+		t.t.Errorf("Error while searching for files in the %s folder: %s", t.folder, err)
+	}
+
+	for _, file := range files {
+		t.t.Run(strings.TrimSuffix(filepath.Base(file), ".php"), func(t *testing.T) {
+			testFileName := file
+			expectedFileName := file + expectedExtension
+			fixedFileName := file + fixExtension
+
+			testFile, err := os.Open(testFileName)
```
quasilyte, line 49: If you're trying to read everything from a file, use `ioutil.ReadFile` instead of Open+ReadAll.
Unless you want a stream-like file handling (but you still do ReadAll right after the close, so it's not the case).

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,126 @@
...
+			var expectedFileContent []byte
+			if expectedFileFound {
+				expectedFileContent, err = ioutil.ReadAll(expectedFile)
+				if err != nil {
+					t.Errorf("Reading file %s failed: %s", expectedFileName, err)
+				}
+				expectedFile.Close()
+			}
+
+			fixedFile, err := os.Open(fixedFileName)
```
quasilyte, line 79: Doing open + close without defer in a big function begs for the resource leak.
I would recommend moving file writing into a separate function, so you can open+defer close+write.
This way, you're guaranteed (more or less) not to leak a file handle.
Also, consider using `iotuil.WriteFile` where possible.
If you don't have to open the file manually, don't do it, so you don't have to close it.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			_, _ = fixedFile.Write(testFileContent)
+			fixedFile.Close()
+
+			test := linttest.NewSuite(t)
+			test.ApplyQuickFixes = true
+			test.AddNamedFile(fixedFileName, string(testFileContent))
+			_ = test.RunLinter()
+
+			fixedFile, err = os.Open(fixedFileName)
```
quasilyte, line 94: I won't comment every line like this, but I do believe you can use `ioutil.ReadFile` here as well. Please re-check.

quasilyte (changes requested):
The implementation could be improved. :nerd_face:

quasilyte:
ioutil writefile requires file permissions, so I wonder if we can copy the original file permissions correctly. But it's OK to keep open+wirte+close, it's just more reliable if moved to a function, so you can defer your close.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+func newQuickFixTest(t *testing.T, folder string) quickFixTest {
+	return quickFixTest{
+		t:      t,
+		folder: folder,
+	}
+}
+
+func openFile(filename string) (f *os.File, found bool, err error) {
+	f, err = os.Open(filename)
```
quasilyte, line 40: You can use other Open call that creates a file if it doesn't exist. I'm not 100% sure why the hard way is used here.

quasilyte: Also, you do open+write+close without extra logic, at least it looks like it.
writeFile can be a function, not openFile. This way your write is more atomic and close can be deferred.
Avoid passing resources that need to be closed when  it's not necessary.

i582: We need to know if the `.fix.expected` file exists in order to implement the creation of a ready-made file with the expected code if it does not exist.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		return f, false, nil
+	}
+	return f, true, nil
+}
+
+func (t *quickFixTest) runQuickFixTest() {
+	files, err := linttest.FindPHPFiles(t.folder)
+	if err != nil {
+		t.t.Errorf("Error while searching for files in the %s folder: %s", t.folder, err)
```
quasilyte, line 54: I would thing that this needs to be a `t.Fatalf` or we need to do a return after Errorf.
Should this test continue if we failed to get a files list?

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,123 @@
...
+				if err != nil {
+					t.Errorf("Reading file %s failed: %s", expectedFileName, err)
+				}
+			}
+
+			fixedFile, _, err := openFile(fixedFileName)
+			if err != nil {
+				t.Errorf("File open %s failed: %s", fixedFileName, err)
+			}
+			_, _ = fixedFile.Write(testFileContent)
```
quasilyte, line 86: If it's intended to ignore the Write error here please clarify why in a comment. Otherwise you need to check for error and do `t.Errorf` in case it's not nil.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,123 @@
...
+			}()
+			_ = test.RunLinter()
+
+			fixedFileContent, err := ioutil.ReadFile(fixedFileName)
+			if err != nil {
+				t.Errorf("Reading file %s failed: %s", fixedFileName, err)
+			}
+
+			if !expectedFileFound {
+				_, _ = expectedFile.Write(fixedFileContent)
```
quasilyte, line 103: Same as above. Error is ignored.

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,123 @@
...
+				t.Logf("The expected files for \"%s\" were not found and were generated automatically.", filepath.Base(testFileName))
+				return
+			}
+
+			expected := string(expectedFileContent)
+			want := string(fixedFileContent)
```
quasilyte, line 109: Why do we have 2 synonyms for want/expected here?
It's usually either have+want or actual+expected.
We essentially have `want+want` here. :sweat_smile: 
I propose renaming `want` to `have` (and `expected` to `want` for consistency).

src/tests/golden/quickfix_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+			if !cmp.Equal(expected, want) {
```
quasilyte, line 111: Both `expected` and `want` (hah) are strings.
Why not just do a `expected != want` comparison?
Or you can do a diff and check it for `""` as in some other tests.

quasilyte (approved):

quasilyte (approved):

PR #731, opened 2020-08-31
linter: added handling of some assignment operators

### LINKED ISSUE #720: Erroneous type inference when using operators like +=
## Code Example

```php
function foo() {
    $a = 0;
    $a += 12.56;
}
```

## Actual Behavior
`$a` is `int`

## Expected Behavior
`$a` is `float`

Most likely the matter is that we do not handle assignment operators like `+=`.

i582:
`+=`
`-=`
`*=`
`.=`

Fixes #720


src/linter/block.go
```diff
@@ -1773,6 +1781,51 @@ func (b *BlockWalker) handleAssign(a *ir.Assign) bool {
...
+	// We are converting "$x += $y" to "$x = $x + $y"
+	// to reuse existing processing code.
+	switch s := o.(type) {
+	case *ir.AssignPlus:
+		as := ir.Assign{
+			Variable: s.Variable,
+			Expression: &ir.PlusExpr{
+				Left:  s.Variable,
+				Right: s.Expression,
+			},
```
quasilyte, line 1794: I think it's better to set positions info here.

Otherwise, when the walker will want to emit a warning it'll give it for some undefined position.

src/linter/block.go
```diff
@@ -1782,45 +1782,47 @@ func (b *BlockWalker) handleAssign(a *ir.Assign) bool {
...
-		as.Walk(b)
+		s.Variable.Walk(b)
+		s.Expression.Walk(b)
+		as := &ir.MinusExpr{
+			Position: s.Position,
+			Left:     s.Variable,
+			Right:    s.Expression,
+		}
+		tp := solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, as)
+		b.replaceVar(s.Variable, tp, "", meta.VarAlwaysDefined)
```
quasilyte, line 1805: Don't leave the reason empty.
You can probably move `b.replaceVar` below the switch statement if all cases end with the same code path.
`tp` variable will go above the switch.

quasilyte (approved):

src/linter/block.go
```diff
@@ -1775,6 +1783,60 @@ func (b *BlockWalker) handleAssign(a *ir.Assign) bool {
...
+			Left:     s.Variable,
+			Right:    s.Expression,
+		}
+		vr = s.Variable
+		tp = solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, as)
+	case *ir.AssignMinus:
+		s.Variable.Walk(b)
+		s.Expression.Walk(b)
+		as := &ir.MinusExpr{
+			Position: s.Position,
```
quasilyte, line 1805: Now that we don't Walk on created nodes, setting positions is redundant.

src/linter/block.go
```diff
@@ -1775,6 +1783,60 @@ func (b *BlockWalker) handleAssign(a *ir.Assign) bool {
...
 
+func (b *BlockWalker) handleAssignOp(o ir.Node) (res bool) {
+	var tp meta.TypesMap
+	var vr ir.Node
+
+	switch s := o.(type) {
+	case *ir.AssignPlus:
+		s.Variable.Walk(b)
+		s.Expression.Walk(b)
+		as := &ir.PlusExpr{
```
quasilyte, line 1794: This vairable is not assignment now. `as` is not the best name for it.

src/linter/block.go
```diff
@@ -1775,6 +1783,60 @@ func (b *BlockWalker) handleAssign(a *ir.Assign) bool {
 	return false
 }
 
+func (b *BlockWalker) handleAssignOp(o ir.Node) (res bool) {
```
quasilyte, line 1786: `o` is probably not the best name for the assignment node.
It's assignment+op, but the main part is assignment, not op.

src/linter/block.go
```diff
@@ -1775,6 +1783,60 @@ func (b *BlockWalker) handleAssign(a *ir.Assign) bool {
...
+		}
+		vr = s.Variable
+		tp = solver.ExprTypeLocal(b.ctx.sc, b.r.ctx.st, as)
+	default:
+		return false
+	}
+
+	b.replaceVar(vr, tp, "assign", meta.VarAlwaysDefined)
+
+	return false
```
quasilyte, line 1837: I think it's easier not to return `false` and don't walk var/expr in adhoc way. It's not necessary for these kinds of nodes.

PR #734, opened 2020-08-31
linter: added Quickfix flag for CheckInfo

### LINKED ISSUE #716: Add Quickfix:bool flag to the CheckInfo
This flag tell whether this checker supports quick fixes or not.

Something like this:
```

	// Quickfix tells whether this checker can automatically fix the reported
	// issues when linter works in -fix mode.
	Quickfix bool

```

i582:
This flag tells whether this checker supports quick fixes or not.

Fixes #716


quasilyte (changes requested):
Please set Quickfix=true for dynamic rules that have a quickfix.

src/cmd/embeddedrules/rules.php
```diff
@@ -9,6 +9,7 @@
  * @comment Report ternary expressions that can be simplified.
  * @before  $x ? $x : $y
  * @after   $x ?: $y
+ * @fix     $x ?: $y
```
quasilyte, line 12: `$x` should be pure. You can't remove duplicated expression if it may have side effects.

quasilyte (approved):

quasilyte:
Nicely done!

PR #743, opened 2020-09-01
linter: added handling of in_array and array_search functions

i582:
If the first argument is of type string, then the third argument is required for correctness.


quasilyte:
I think there is a typo in the commit message: it should be `in_array`, not `is_array`.
GitHub doesn't allow to comment on commit messages (Gerrit does).

src/cmd/embeddedrules/rules.php
```diff
@@ -238,3 +238,52 @@ function argsOrder() {
...
+        false == $_;
+        $_ == null;
+        null == $_;
+        $x == $y;
+    }
+
+    /**
+     * @warning non-strict comparison (use !==)
+     * @type string $x
+     * @type string $y
```
quasilyte, line 266: I'm not sure what these type filters are supposed to do.
Maybe we can remove them?
Patterns below don't have `$x` or `$y` matcher variables.
I would guess it's a copy/paste error.

src/cmd/embeddedrules/rules.php
```diff
@@ -238,3 +238,52 @@ function argsOrder() {
...
+/**
+ * @comment Report not-strict-enough comparisons.
+ * @before  in_array("what", $s)
+ * @after   in_array("what", $s, true)
+ */
+function strictCmp() {
+    /**
+     * @warning non-strict comparison (use ===)
+     * @type string $x
+     * @type string $y
```
quasilyte, line 251: Same as below. Please remove them if you don't need them.

quasilyte:
If conflicts are solved and this rule was tested on the vkcom, then it can be merged.

i582:
Closed due #830

PR #744, opened 2020-09-01
ir/irfmt: fix array item expr printing with unpack=true

quasilyte:
See https://github.com/z7zmey/php-parser/pull/116


quasilyte:
We also had swapped have/want arguments diff/equal calls, so I fixed that as well.

PR #752, opened 2020-09-02
linter: fix issues in typeIsCompatible() func

quasilyte:
This function is only used in dynamic rules for the
@type filter check.

Also added tests to verify that we get the basic things right.


quasilyte:
Since this issue is broader than I thought, I'll try to cover #753 in one way or another.

PR #771, opened 2020-09-07
tests: make e2e tests ~x6 faster

quasilyte:
Instead of doing a full stubs loading, use our dependencies
list from the golden tests to hint the linter which embedded
stubs need to be loaded.

With this, we significantly reduce the amount of work linter
needs to do before running the inspections.

On my machine, golden tests run for 40-50 seconds.
After this patch I get 5-8 seconds.

### DIFF
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -351,7 +351,14 @@ func LoadEmbeddedStubs(filenames []string) error {
 }
 
 func loadEmbeddedStubs() error {
-	filenames := stubs.AssetNames()
+	var filenames []string
+	// NOVERIFYDEBUG_LOAD_STUBS is used in golden tests to specify
+	// the test dependencies that need to be loaded.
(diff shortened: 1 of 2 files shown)


i582 (approved):

YuriyNasretdinov:
I would recommend to introduce a separate flag instead, e.g. "-test-only-stubs-list". Environment variables are a convenient thing but maintaining the code that uses environment variables, especially to hide some functionality, is harder in longer term. Think of it as a global variable, but that is also implicit.

YuriyNasretdinov:
Alternatively, maybe call the environment variable something like "NOVERIFY_TEST_LOAD_STUBS", so that it can be easily found in the codebase. Any way to make it more explicit without having users to worry about it would be good in my opinion.

YuriyNasretdinov (approved):

PR #794, opened 2020-09-22
linttest: all functions for golden testing have been moved to a separate framework golden_linttest

i582:
The main functions remain the same, only the wrapper has changed.
This change will allow us to run golden tests in projects built on NoVerify.


src/linttest/linttest.go
```diff
@@ -79,7 +82,7 @@ type Suite struct {
 	AllowDisable *regexp.Regexp
 
 	defaultStubs map[string]struct{}
-	LoadStubs    []string
+	LoadedStubs  []string
```
quasilyte, line 85: It's not obvious why this renaming is needed.
I don't know why we need a setter method for it as well.

src/linttest/linttest.go
```diff
@@ -116,6 +119,27 @@ func (s *Suite) AddNamedFile(name, contents string) {
...
+	}
+}
+
+// LoadStubs adds stubs to the test.
+func (s *Suite) LoadStubs(stubs []string) {
+	s.LoadedStubs = stubs
+}
+
+// SetMisspellList set value of MisspellList.
+func (s *Suite) SetMisspellList(name string) {
```
quasilyte, line 139: Why adding a setter?

i582: Deleted

src/linttest/linttest.go
i582: Deleted

src/linttest/golden_linttest.go
```diff
@@ -0,0 +1,302 @@
...
+
+// PrepareGoldenTestSuite configures fields and standard stubs.
+//
+// Used if the structure was created directly.
+func PrepareGoldenTestSuite(s *GoldenTestSuite, t *testing.T, baseDir, goldenFileName string) {
+	s.suite = NewSuite(t)
+	s.BaseDir = baseDir
+	s.prepared = true
+	s.GoldenFileName = goldenFileName
+	s.Deps = append(s.Deps, defaultStubs...)
```
quasilyte, line 83: Maybe use AddDeps here?

PR #795, opened 2020-09-22
linttest: removed unnecessary AddNamedFile function

i582:
Because there is an `AddNamedFile` method


quasilyte (approved):

YuriyNasretdinov:
Did you try to sleep at night?

PR #797, opened 2020-09-24
linter: utils moved to a separate folder

i582:
We need to get rid of the fully dependency on the linter package
in order to be able to refactor the code easier in the future
and not get cycle import errors.

### DIFF
--- a/src/linter/and_walker.go
+++ b/src/linter/and_walker.go
@@ -2,6 +2,7 @@ package linter
 
 import (
 	"github.com/VKCOM/noverify/src/ir"
+	"github.com/VKCOM/noverify/src/linter/utils"
 	"github.com/VKCOM/noverify/src/meta"
 	"github.com/VKCOM/noverify/src/solver"
 )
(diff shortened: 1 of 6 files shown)


YuriyNasretdinov:
Generally and especially in Go calling your package "utils" and such is frowned upon as this package name literally doesn't say anything about what it is for. Having big packages when you can't split into non-arbitrary packages is actually OK generally.

i582:
But after all, when everything is intertwined in a package, it is bad, and placing a part in a separate package allows us to reduce this intertwining, isn't it?

YuriyNasretdinov:
Not sure that your change makes it less intertwined honestly

quasilyte:
I'm not 100% against util packages when they have some meaning attached to them, like `astutil`, but it's even better when you can call a package based on the main functionality. `solver` is really is a helper package to resolve the types, but it's not called `typeutils`, it's called `solver`. If we can't give a good name for things inside a package, it's probably not a good idea to move things there.

PR #800, opened 2020-09-29
all: fix golang-ci

i582:


i582:
Closed due to golang-ci problem resolving itself.

PR #802, opened 2020-10-01
constfold: added support for magic constants in constant folding, as …

i582:
…well as the `dirname(__FILE__)` function

### DIFF
--- a/src/constfold/eval.go
+++ b/src/constfold/eval.go
@@ -1,6 +1,7 @@
 package constfold
 
 import (
+	"path/filepath"
 	"strconv"
 
 	"github.com/VKCOM/noverify/src/ir"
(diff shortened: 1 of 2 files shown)


src/constfold/eval.go
```diff
@@ -94,6 +95,39 @@ func Eval(st *meta.ClassParseState, e ir.Node) meta.ConstValue {
 
 	case *ir.String:
 		return meta.ConstValue{Value: e.Value, Type: meta.String}
+
+	case *ir.FunctionCallExpr:
+		// dirname(__FILE__)
+		if !meta.NameNodeEquals(e.Function, `dirname`) {
+			return meta.UnknownValue
+		}
+		arg, ok := e.Arg(0).Expr.(*ir.MagicConstant)
```
quasilyte, line 104: Won't this code panic if dirname is called with 0 args?

Like `dirname()`.

quasilyte (changes requested):
Please add some tests.
I think you can use normalizer package tests for that.

quasilyte (approved):

PR #807, opened 2020-10-06
linter: added support for @mixin annotation

### LINKED ISSUE #782: Add support for @mixin annotation
## Code Example

```php
<?php

class Foo {
    public function method() {}
}

/**
 * Class Bar
 * @mixin Foo
 */
class Bar {
    public function run()
    {
        $this->method();
    }
}
```

## Actual Behavior

```
<critical> ERROR   undefined: Call to undefined method {\Bar}->method() at Sample.php:14
        $this->method();
               ^^^^^^
```

## Expected Behavior

No errors found

i582:
Fixes #782

### DIFF
--- a/src/linter/phpdoc_util.go
+++ b/src/linter/phpdoc_util.go
@@ -113,6 +113,9 @@ func parseClassPHPDocMixin(ctx *rootContext, result *classPhpDocParseResult, par
 	if !strings.HasPrefix(param, `\`) {
 		param = `\` + param
 	}
+	if ctx.st.Namespace != "" {
+		param = ctx.st.Namespace + param
+	}
 
 	result.mixins = append(result.mixins, param)
 }
--- a/src/linter/phpdoc_util.go
+++ b/src/linter/phpdoc_util.go
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/VKCOM/noverify/src/ir"
 	"github.com/VKCOM/noverify/src/meta"
 	"github.com/VKCOM/noverify/src/phpdoc"
+	"github.com/VKCOM/noverify/src/solver"
 )
 
 type phpdocErrors struct {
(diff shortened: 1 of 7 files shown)


src/linter/phpdoc_util.go
```diff
@@ -100,3 +101,18 @@ func parseClassPHPDocProperty(ctx *rootContext, result *classPhpDocParseResult,
...
+
+func parseClassPHPDocMixin(ctx *rootContext, result *classPhpDocParseResult, part *phpdoc.RawCommentPart) {
+	params := part.Params
+
+	if len(params) == 0 {
+		return
+	}
+
+	param := params[0]
+	if !strings.HasPrefix(param, `\`) {
```
YuriyNasretdinov, line 113: Usually the classes mentioned in phpdoc are defined inside a namespace. You need to resolve the class name the same way it is implemented for @return types, for example.

src/linter/root.go
```diff
@@ -1263,6 +1271,27 @@ func (d *RootWalker) checkPHPDocRef(n ir.Node, part phpdoc.CommentPart) {
...
+
+	if len(params) == 0 {
+		return
+	}
+
+	param := params[0]
+	if !strings.HasPrefix(param, `\`) {
+		param = `\` + param
+	}
+	if _, ok := meta.Info.GetClass(param); !ok {
```
YuriyNasretdinov, line 1290: Are you sure that you finished indexing here?

YuriyNasretdinov (changes requested):
1. Namespaces must be supported
2. Not 100% sure (I might be totally wrong) whether or not you are trying to read things from the index before indexing is complete.

src/linter/phpdoc_util.go
i582: Sorry if I misunderstood you. I tried testing the following:

```php
// file1.php
namespace QQ;

class SomeQQClass {
  /** */
  public function methodQQ()
  {
    echo "";
  }
}
```
```php
// file2.php
/**
 * @mixin QQ\SomeQQClass
 */
class SomeClass {
  /** */
  public function method()
  {
    echo $this->methodQQ();
  }
}
```

This works correctly, there are no errors.

Did you mean that?

YuriyNasretdinov: You can also write the following thing, for example:

```
<?php
namespace QQ;

/** @mixin SomeQQClass */
class A {
}
```

SomeQQClass here actually means \QQ\SomeQQClass

i582: Thank you I got it.

YuriyNasretdinov: That was only one example :). I suggest that you use the existing functions that resolve types in phpdoc instead. There are more cases that should be properly handled, e.g.

```
<?php

namespace QQ;

use A as B;

class C {
/** @mixin A */
// This actually means mixin \QQ\B
}
```

YuriyNasretdinov (approved):

PR #816, opened 2020-11-17
docs: update docs

i582:


docs/vscode-plugin.md
```diff
@@ -12,6 +12,7 @@ For example, you can configure the following after VS Code installation:
 {
   "php-noverify.noverifyPath": "<noverify binary path>",
   "php-noverify.noverifyExtraArgs": [
+    "check",
     "-cores=4",
     "-cache-dir=/path/to/cache"
```
YuriyNasretdinov, line 17: You don't need a cache-dir here anymore

docs/writing-new-ide-plugin.md
```diff
@@ -5,7 +5,7 @@ NoVerify implements [Language Server Protocol](https://langserver.org) for PHP,
 Use the following command to run `noverify` as language server:
 
 ```sh
-$ noverify -lang-server -cores=4 -cache-dir=/path/to/cache
+$ noverify check -lang-server -cores=4 -cache-dir=/path/to/cache
```
YuriyNasretdinov, line 8: no need to specify cache dir anymore

YuriyNasretdinov (approved):

docs/linter-usage.md
```diff
@@ -5,10 +5,10 @@
 In order to get reports for all files in repository, run the following:
 
 ```sh
-$ noverify -cache-dir=$HOME/tmp/cache/noverify /path/to/your/project/root
+$ noverify check [-cache-dir=$HOME/tmp/cache/noverify] /path/to/your/project/root
```
quasilyte, line 8: I don't think that `[` and `]` are intuitive.
Usually, these snippets are interpreted as something you can copy/paste and use right away.
You can make it 2 examples: 1 with default cache and another with a custom location.
Comments can help to make it obvious why you may need to use a custom cache location and what's the behavior of the default one.

docs/linter-usage.md
```diff
@@ -5,10 +5,10 @@
...
 ```
 
-Cache dir is optional, but recommended. Next launch would be much faster with cache if you specify some cache directory.
+The cache directory is optional, by default it is already set to `temporary_folder_path/noverify`, but if you want to change the folder for storing the cache, then use the `-cache-dir` flag.
```
quasilyte, line 11: I would write something like `$TMPDIR/noverify`. Otherwise it looks like `temporary_folder_path` is some kind of the literal path.

quasilyte (approved):

PR #819, opened 2020-11-23
add check nullable syntax

i582:
Before that, if a `T?` entry appeared in `phpdoc`, the type was always deduced as null, which led to false positives. Now a warning will be issued for such a record.


quasilyte (changes requested):
Please add tests for shape types with optional keys.

src/linter/types.go
```diff
@@ -101,6 +102,13 @@ func (conv *phpdocTypeConverter) mapType(e phpdoc.TypeExpr) []meta.Type {
 			types = append(types, conv.mapType(a)...)
 		}
 		return types
+
+	case phpdoc.ExprOptional:
+		// If the type does not denote the shape or tuple type,
+		// then it is most likely an invalid nullable type syntax.
+		if !strings.Contains(e.Value, "shape") && !strings.Contains(e.Value, "tuple") {
```
quasilyte, line 109: Tuples can't have optional keys.

quasilyte: Also, I don't think `e.Value` would contain what you're looking for. You probably need to save the context like we do for the nullable types.

quasilyte: ping

PR #826, opened 2020-11-30
linter: added discardVar check

### LINKED ISSUE #825: Add discardVar check
## Code Example

```php
$_ = 100;
echo $_;
```

## Actual Behavior

No errors,

## Expected Behavior

Error using discard variable `$_`.

## Description

Since `$_` var usages are not tracked, warn if someone ends up using them.
    
It's especially possible when to discard vars are signified by `$_` prefix, so `$_id` might be used. In that case, the user should rename the variable to `$id` or stop using it.

i582:
Fixes #825

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1366,6 +1366,9 @@ func (b *BlockWalker) handleVariable(v ir.Node) bool {
 				w.untrackVarName(varName)
 			}
 		}
+		if IsDiscardVar(varName) && !isSuperGlobal(varName) {
+			b.r.Report(v, LevelError, "discardVar", "Used discard variable $%s value (rename variable if it's intended)", varName)
+		}
 
 		b.untrackVarName(varName)
 	}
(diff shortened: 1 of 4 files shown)


YuriyNasretdinov (approved):

quasilyte (changes requested):
Please add tests for cases that only work correctly due to the implementation detail.
If we ever split checkers and will run them separately, this code might break otherwise without us noticing it.

* list args - `list ($_unused, ...) = $rhs`; `$_unused` is not a usage
* lambda params - `$f = function($_unused, $x) {...}`; `$_unused` is not a usage
* same for arrow function params

There might be other interesting cases.

src/linter/block.go
```diff
@@ -1366,6 +1366,9 @@ func (b *BlockWalker) handleVariable(v ir.Node) bool {
 				w.untrackVarName(varName)
 			}
 		}
+		if IsDiscardVar(varName) && !isSuperGlobal(varName) {
+			b.r.Report(v, LevelError, "discardVar", "Used discard variable $%s value (rename variable if it's intended)", varName)
```
quasilyte, line 1370: "discard value" is our own term which won't be too obvious for the users.

How about this message:
```
Used var that is supposed to be unused (rename variable if it's intended)
```

quasilyte (approved):

PR #828, opened 2020-12-03
add support of is_array

i582:
Now, in bodies with the `is_array` condition, the type will be inferred as follows:

- If the variable has only types that represent arrays, then the type does not change.
```php
  int[]|Foo[] -> int[]|Foo[]
```

- If the variable has at least one type that does not represent an array, then the body
  will contain only types representing arrays.
```php
  int|Foo[] -> Foo[]
```

- If the variable has no types that represent arrays, then the type in the body will be `mixed[]`.
```php
  int|string -> mixed[]
```

In order to restore the original types of a variable after leaving the body, the `andWalker.varsToReplace` 
field has been added, which stores an array of variables and their types to which must be replaced after
leaving the body.


src/meta/typesmap.go
```diff
@@ -173,6 +173,34 @@ func (m TypesMap) IsArray() bool {
 	return false
 }
 
+// HasOnlyArrays checks if map contains only arrays of any type
+func (m TypesMap) HasOnlyArrays() bool {
```
quasilyte, line 177: These 2 new TypesMap predicates only used once. Why not use `m.Contains` with a callback there?
TypesMap already has quite a lot of such 1-time use (very ad-hoc) methods.
If you really want them, you can write them as util functions somewhere in linter package.
The simpler TypesMap API is, the easier it will be to refactor/replace it later.

i582: Done.
The functions were inlined.

i582:
Closed due #1038.

PR #838, opened 2020-12-07
WIP: linter: handle class name case in const fetch

quasilyte:


YuriyNasretdinov (approved):

quasilyte:
Interesting. Looks like there are some problems with `self` and other special class names. I'll look into that a little bit later.

i582:
Closed due #1046

PR #843, opened 2020-12-09
solver: added type inference for KPHP-specific functions array_last_e…

i582:
…lement and array_first_element


src/meta/metainfo.go
```diff
@@ -157,6 +183,8 @@ func (i *info) InitStubs() {
 	for k, v := range i.allFunctionsOverrides {
 		internalFunctionOverrides[k] = v
 	}
+
+	i.InitKphpStubs()
```
quasilyte, line 187: I think this code should be executed only if we have `-kphp` flag set to true.

src/meta/metainfo.go
```diff
@@ -131,6 +131,32 @@ func (i *info) FindConstants(substr string) (res []string) {
 	return res
 }
 
+func (i *info) InitKphpStubs() {
```
quasilyte, line 134: Can this function be moved to `linter` package so it can check the `kphp` flag?
Or we can pass it to the Init stubs function to avoid redundant meta overrides for non-kphp mode.

quasilyte (approved):
LGTM.

PR #844, opened 2020-12-09
WIP: config: move conf.go to config package

i582:
This is necessary so that you can use the KPHP variable when initializing stubs.


i582:
Not the best way.

Issue #849, opened 2020-12-10
Experiment: alternative function result types inference


quasilyte:
To make function result types inference more predictable and controllable, we'll try to restrict it.

New function result type inference rules:

* If `@return` annotation exists, use that.
* Otherwise use return type hint.
* If neither present, infer actually returned types.

```php
// @return is present, f1() -> int[]
/** @return int[] */
function f1(): array { ... }

// type hint is present, f2() -> mixed[]
function f2(): array { ... }

// type info is infered from actual results, f3() => int[]
function f3() { return [1, 2]; }
```

If we've used `@return` or a type hint, we need to perform a `typecheckReturn`
and verify that actually returned values do not contradict stated types.

Instead of doing a full value (and types) analysis for `typecheckReturn`, it's
proposed to perform a set of checks that detect only critical issues.

This set can be extended as we adopt this approach. At some point,
we may switch to a full types analysis and handle it as PHP would.

Examples of the critical issues we should report in `typecheckReturn`:

* Array types mismatch
  * @return array, but returned non-array
  * @return !array, but returned array
* Primitive/scalar types returned in place of object
  * @return object-type, but returned scalar type
  * @return scalar-type, but returned object type

Later, we may add checks for nullability.

It's also possible to check interface-returned types and whether
one class may be returned in place of another; see [Covariance and Contravariance](https://www.php.net/manual/en/language.oop5.variance.php).

But since that's more complicated and we need to start somewhere and prepare the foundation,
let's focus on the simple things first and test it on the real projects.

Note: we should be ready to revert our changes in case this experiment is proven to be a failure.

i582:
I consider the experiment to be successful.
Further type checks will be tracked in #1123.

PR #851, opened 2020-12-14
linter: added a new way of return type inference for functions and methods

i582:
The types are now inferred as follows:
1. If there is an `@param` annotation, then its value becomes
   the return type;

2. If there is a type hint, then it is added to the types
   from the `@param`. If the `@param` is empty, then the type
   matches the type hint itself;

3. If there is no `@param` annotation and type hint, then
   the return type is equal to the union of the types that
   are returned from the function by return.

Part of #849 

The next step, adding type checking.


src/linter/utils.go
```diff
@@ -90,6 +90,32 @@ func typesMapToTypeExpr(p *phpdoc.TypeParser, m meta.TypesMap) phpdoc.Type {
 	return p.Parse(typeString)
 }
 
+// functionReturnType returns the return type of a function over computed types
+// according to the convention below:
+//
+// The types are inferred as follows:
+// 1. If there is an @param annotation, then its value becomes the return type;
```
quasilyte, line 97: Maybe `@return` annotation was implied here (and below)?

src/linter/utils.go
```diff
@@ -90,6 +90,32 @@ func typesMapToTypeExpr(p *phpdoc.TypeParser, m meta.TypesMap) phpdoc.Type {
...
+//
+// 2. If there is a type hint, then it is added to the types from the @param.
+//    If the @param is empty, then the type matches the type hint itself;
```
quasilyte, line 100: Does merging type hint and `@return` improve the type inference?

For example, in new tests there is `f6()` which has `array` type hint and `int[]` from phpdoc annotation.
I would guess that marking `f6()` result as `int[]` is more desirable than `int[]|mixed[]`.

src/linter/utils.go
i582: Yep. Done.

quasilyte:
1. Please increase the cache version as we changed the way how function results are inferred.
2. Add something in a cache_test so we get a different result for the type inference before and after this change.

I believe the cache_test is not failing because we don't have a function there that would be inferred differently.

src/linter/utils.go
i582: Done

quasilyte (approved):

PR #855, opened 2020-12-14
cmd: added a message "Some issues are autofixable (try using the `-fi…

### LINKED ISSUE #845: Print "some issues are autofixable" if any of reported issues supports autofix
To make `-fix` feature more discoverable, we can print a teaser like this:

```
2020/12/10 13:37:22.612558 Linting
MAYBE   arraySyntax: Use of old array syntax (use short form instead) at example.php:3
var_dump(array(1, 2));
         ^^^^^^^^^^^
2020/12/10 13:37:22.612558 Some issues are autofixable (try using the `-fix` flag)
```

i582:
…x` flag)" if any error can be automatically fixed

Fixes #845


src/cmd/main.go
```diff
@@ -319,7 +323,30 @@ func FormatReport(r *linter.Report) string {
...
+			continue
+		}
+
+		checksWithQuickfix[check.Name] = struct{}{}
+	}
+
+	var withAutofixableReports bool
+	for _, r := range reports {
+		if _, ok := checksWithQuickfix[r.CheckName]; ok {
+			withAutofixableReports = true
```
quasilyte, line 341: Why not just `return true` here and `return false` in the end of the function?

quasilyte (approved):

PR #859, opened 2020-12-14
solver: added initial support for instance_deserialize in KPHP mode

i582:
At the moment, string assignment of the class
and assignment through the class constant are
supported:

```php
instance_deserialize ("", Foo::class);
instance_deserialize ("", "Foo");
```


src/tests/exprtype/exprtype_test.go
```diff
@@ -2707,6 +2707,32 @@ function f1() {
 	runExprTypeTest(t, &exprTypeTestParams{code: code, stubs: "<?php /* no code */"})
 }
 
+func TestInstanceDeserializeType(t *testing.T) {
```
quasilyte, line 2710: Consider to move this test to `kphp_test.go`.

i582: Done

quasilyte (approved):
Looks legit.

src/solver/exprtype.go
```diff
@@ -202,6 +202,41 @@ func binaryPlusOpType(sc *meta.Scope, cs *meta.ClassParseState, left, right ir.N
...
+		default:
+			return "", false
+		}
+	default:
+		return "", false
+	}
+
+	className, ok := GetClassName(cs, &ir.Name{
+		Value: name,
+	})
```
quasilyte, line 232: Can this literal be placed on the same line as GetClassName call?
```go
	className, ok := GetClassName(cs, &ir.Name{Value: name})
```

i582: Done

quasilyte (approved):

PR #862, opened 2020-12-17
linter: added storage of default values for params of functions and methods

i582:


src/linter/cache_test.go
```diff
@@ -101,6 +101,18 @@ function as_shape($x) {
...
 }
 
+/** With default */
+function withDefault($x = 100) {
+  return $x;
+}
+
+class WithDefaultMethod {
+  /** With default */
+  public function withDefault($x = 100) {
```
quasilyte, line 111: Use different type of the default param here. String, for example.

src/meta/metainfo.go
```diff
@@ -312,9 +312,10 @@ func (i *info) AddToGlobalScopeNonLocked(filename string, sc *Scope) {
 }
 
 type FuncParam struct {
-	IsRef bool
-	Name  string
-	Typ   TypesMap
+	IsRef   bool
+	Name    string
+	Typ     TypesMap
+	Default *ConstValue
```
quasilyte, line 318: Why can't it be a non-pointer type?
Zero value ConstValue has Type=Undefined, which is good enough to detect "no default value" case.

src/linter/cache_test.go
i582: Done.

src/meta/metainfo.go
i582: Good idea. Done.

i582:
Let's leave until better times.

PR #863, opened 2020-12-17
all: initial types for element of TypesMap

i582:
Trying to type the `TypesMap` elements for more convenient interaction and, ahem, more typed.

## Motivation:
Initially, I wanted to avoid low-level interactions, such as here:

```go
func MergeTypeMaps(left meta.TypesMap, right meta.TypesMap) meta.TypesMap {
	var hasAtLeastOneArray bool
	var hasAtLeastOneClass bool
	newMap := meta.NewEmptyTypesMap(left.Len() + right.Len())

	left.Iterate(func(typ string) {
		if typ[0] == meta.WArrayOf {
			hasAtLeastOneArray = true
		}
		if typ[0] == '\\' {
			hasAtLeastOneClass = true
		}
		newMap = newMap.AppendString(typ)
	})

	right.Iterate(func(typ string) {
		if typ[0] == meta.WArrayOf && meta.UnwrapArrayOf(typ) == "mixed" && hasAtLeastOneArray {
			return
		}
		if typ == "object" && hasAtLeastOneClass {
			return
		}
		newMap = newMap.AppendString(typ)
	})

	return newMap
}
```

So that we can conveniently work with individual types without going into implementation details and see all the methods that are available for them, and not a certain set of functions, such as `IsArrayType`, which for some reason is in the `meta` package.

Now we can write this and it looks more high-level and does not depend on the implementation details.

```go
func MergeTypeMaps(left meta.TypesMap, right meta.TypesMap) meta.TypesMap {
	var hasAtLeastOneArray bool
	var hasAtLeastOneClass bool

	res := meta.NewRawTypesMap(left.Len() + right.Len())

	left.Iterate(func(typ meta.Type) {
		if typ.IsArray() {
			hasAtLeastOneArray = true
		}
		if typ.IsClass() {
			hasAtLeastOneClass = true
		}
		res = res.Append(typ)
	})

	right.Iterate(func(typ meta.Type) {
		if typ.IsArray() && typ.ElementType().IsMixed() && hasAtLeastOneArray {
			return
		}
		if typ.Is("object") && hasAtLeastOneClass {
			return
		}
		res = res.Append(typ)
	})

	return meta.NewTypesMapFromMap(res)
}

```


src/linter/cache_test.go
```diff
@@ -135,7 +135,7 @@ main();
 		// 2. Check cache "strings" hash.
 		//
 		// It catches new fields in cached types, field renames and encoding of additional named attributes.
-		wantStrings := "62321eae8ac6f753221d367a0a0e2a3190202ef49c608ceef6e03dd6b09906c64bf20a1470abef9e26b917836a68415a5459491208fb692d8773524cdbe6d239"
+		wantStrings := "dc486a1cdb1bd6dcac2742113c0eeeed28f67644704c0fda1c7229e6620d3c838c39f2cf0ce83736eea9ef23cf8818197ff7ea851f0987fdddefaa6ae26f4ef1"
```
YuriyNasretdinov, line 138: Updating the test but not the cache version... you are not paid to think

YuriyNasretdinov (changes requested):
You can do better

src/langsrv/refsvisitors.go
```diff
@@ -315,8 +315,8 @@ func (d *blockMethodCallVisitor) BeforeEnterNode(n ir.Node) {
 
 	exprType := solver.ExprType(d.ctx.Scope(), d.ctx.ClassParseState(), call.Variable)
 
-	exprType.Iterate(func(typ string) {
-		m, ok := solver.FindMethod(typ, methodName)
+	exprType.Iterate(func(typ meta.Type) {
+		m, ok := solver.FindMethod(typ.String(), methodName)
```
YuriyNasretdinov, line 319: How about making API that accepts types actually accept types and not require this .String() thing? Basically almost every time you call .String() you want the function to accept the meta.Type instead.

src/linter/block.go
```diff
@@ -1514,33 +1518,31 @@ func (b *BlockWalker) handleAssignList(list *ir.ListExpr, rhs ir.Node) {
...
-	var elemTypes []meta.Type
-	var shapeType string
-	typ.Iterate(func(typ string) {
+	var elemTypes = make(meta.RawTypesMap, typ.Len())
+	var shapeType meta.Type
+
+	typ.Iterate(func(typ meta.Type) {
 		switch {
-		case meta.IsShapeType(typ):
+		case typ.IsShape():
```
YuriyNasretdinov, line 1526: So is this basically the problem that you are trying to solve by refactoring everything to concrete types? Why not put that into the PR description?

src/linter/block_linter.go
```diff
@@ -199,12 +199,12 @@ func (b *blockLinter) checkArrayDimFetch(s *ir.ArrayDimFetchExpr) {
...
 
-	typ.Iterate(func(t string) {
+	typ.Iterate(func(t meta.Type) {
 		// FullyQualified class name will have "\" in the beginning
-		if meta.IsClassType(t) {
+		if t.IsClass() {
 			maybeHaveClasses = true
 
-			if !haveArrayAccess && solver.Implements(t, `\ArrayAccess`) {
+			if !haveArrayAccess && solver.Implements(t.String(), `\ArrayAccess`) {
```
YuriyNasretdinov, line 207: Same here. `solver.Implements` just needs to accept the meta.Type argument.

src/linter/root.go
```diff
@@ -1395,8 +1395,8 @@ func (d *RootWalker) parsePHPDoc(n ir.Node, doc []phpdoc.CommentPart, actualPara
 			result.errs.pushType("%s on line %d", warning, part.Line())
 		}
 		param.typ = newTypesMap(&d.ctx, types)
-		param.typ.Iterate(func(t string) {
-			if t == "void" {
+		param.typ.Iterate(func(t meta.Type) {
+			if t.Is("void") {
```
YuriyNasretdinov, line 1399: You may want to introduce `typ.Contains()` method or something like that instead?

src/linter/types.go
```diff
@@ -15,7 +15,7 @@ type warningString string
 
 type shapeTypeProp struct {
 	key   string
-	types []meta.Type
+	types []meta.PhpDocType
```
YuriyNasretdinov, line 18: Or is this the problem that you are trying to solve? Trying to determine whether or not the types come from the code or from PHPDoc?

src/linter/utils.go
```diff
@@ -111,14 +111,17 @@ func resolveFunctionCall(sc *meta.Scope, st *meta.ClassParseState, customTypes [
...
 	} else {
-		solver.ExprTypeCustom(sc, st, call.Function, customTypes).Iterate(func(typ string) {
+		functionType := solver.ExprTypeCustom(sc, st, call.Function, customTypes)
+
+		functionType.Iterate(func(typ meta.Type) {
 			if res.isFound {
 				return
 			}
-			m, ok := solver.FindMethod(typ, `__invoke`)
+			m, ok := solver.FindMethod(typ.String(), `__invoke`)
```
YuriyNasretdinov, line 120: FindMethod needs to accept Type as the first argument.

src/linter/utils.go
```diff
@@ -161,13 +164,13 @@ func resolveMethodCall(sc *meta.Scope, st *meta.ClassParseState, customTypes []s
...
-	methodCallerType.Find(func(typ string) bool {
-		m, isMagic, ok := findMethod(typ, methodName)
+	methodCallerType.Find(func(typ meta.Type) bool {
+		m, isMagic, ok := findMethod(typ.String(), methodName)
 		if !ok {
 			return false
 		}
 		foundMethod = true
-		if dist := classDistance(st, typ); dist < matchDist {
+		if dist := classDistance(st, typ.String()); dist < matchDist {
```
YuriyNasretdinov, line 173: As should classDistance

src/meta/type.go
```diff
@@ -0,0 +1,59 @@
...
+func (t Type) String() string {
+	return string(t)
+}
+
+func (t Type) IsEmpty() bool {
+	return t == ""
+}
+
+func (t Type) ElementType() Type {
+	if t[0] == WArrayOf {
```
YuriyNasretdinov, line 30: The string can be empty.

src/meta/type.go
```diff
@@ -0,0 +1,59 @@
...
+
+	return NewType(strings.TrimSuffix(string(t), "[]"))
+}
+
+func (t Type) Is(str string) bool {
+	return t.String() == str
+}
+
+func (t Type) IsLazy() bool {
+	return t[0] < WMax
```
YuriyNasretdinov, line 42: Once again, the string can be empty.

src/meta/type.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (t Type) IsMixed() bool {
+	return t == "mixed"
+}
+
+func (t Type) IsArray() bool {
+	return strings.HasSuffix(string(t), "[]") || t[0] == WArrayOf
```
YuriyNasretdinov, line 50: And once more, the string can be empty I said

src/meta/typesmap.go
```diff
@@ -30,16 +30,30 @@ const (
 	mapPrecise
 )
 
+type RawTypesMap map[Type]struct{}
```
YuriyNasretdinov, line 33: Zero value cannot be used directly, you better provide the New() function in this case.

src/meta/typesmap.go
```diff
@@ -72,30 +86,32 @@ func (m TypesMap) IsResolved() bool {
...
-func NewTypesMapFromTypes(types []Type) TypesMap {
-	m := make(map[string]struct{}, len(types))
-	for _, typ := range types {
-		s := typ.Elem
-		for i := 0; i < typ.Dims; i++ {
-			s = WrapArrayOf(s)
-		}
-		m[s] = struct{}{}
+func NewTypesMapFromPhpDocTypes(phpDocTypes []PhpDocType) TypesMap {
+	m := make(RawTypesMap, len(phpDocTypes))
```
YuriyNasretdinov, line 97: I would suggest to introduce New() method here instead of making an arbitrary type...

src/meta/typesmap.go
```diff
@@ -72,30 +86,32 @@ func (m TypesMap) IsResolved() bool {
...
-			s = WrapArrayOf(strings.TrimSuffix(s, "[]"))
+	m := make(RawTypesMap, strings.Count(str, "|")+1)
+
+	for _, typeStr := range strings.Split(str, "|") {
+		typ := NewType(typeStr)
+		if typ.IsArray() {
+			typ = WrapArrayOf(typ.ElementType())
 		}
-		m[s] = struct{}{}
+		m.Add(typ)
```
YuriyNasretdinov, line 116: I don't think it is really obvious when you are adding something to a value type (it is RawTypesMap, it is not *RawTypesMap) does this mean that the value will change? That is the reason why TypesMap provides `Append()` functions that return you (potentially a new) value.

src/meta/typesmap.go
```diff
@@ -193,19 +244,42 @@ func (m TypesMap) Is(typ string) bool {
...
+	}
+
+	mm := make(RawTypesMap, 1)
+	for k, v := range m.m {
+		mm[k] = v
+	}
+
+	mm[typ] = struct{}{}
+
+	// The returned map is mutable and imprecise.
```
YuriyNasretdinov, line 274: Why is it imprecise? If you are appending the precise Type, how do you mark it as such?

src/meta/typesmap.go
```diff
@@ -193,19 +244,42 @@ func (m TypesMap) Is(typ string) bool {
...
+	return TypesMap{m: mm}
+}
+
 // AppendString adds provided types to current map and returns new one (immutable maps are always copied)
 func (m TypesMap) AppendString(str string) TypesMap {
 	if !m.isImmutable() {
 		if m.m == nil {
-			m.m = make(map[string]struct{}, strings.Count(str, "|")+1)
+			m.m = make(RawTypesMap, strings.Count(str, "|")+1)
```
YuriyNasretdinov, line 282: I think a `New()` function would work better here.

src/meta/typesmap.go
```diff
@@ -384,9 +463,9 @@ func (m TypesMap) ArrayElemLazyType() TypesMap {
 		return MixedType
 	}
 
-	mm := make(map[string]struct{}, m.Len())
+	res := make(RawTypesMap, m.Len())
 	for typ := range m.m {
-		mm[UnwrapArrayOf(typ)] = struct{}{}
+		res.Add(typ.UnwrapArrayOf())
```
YuriyNasretdinov, line 472: Please replace it with an append-like API

src/meta/typeutils.go
```diff
@@ -120,204 +120,205 @@ func wrap(typ byte, byteFields []uint8, args ...string) string {
...
 
-func unwrap3(s string) (b1 uint8, one, two string) {
+func (t Type) unwrap3() (b1 uint8, one, two string) {
 	var l int
 	var b [stringLenBytes]byte
 	var rawBuf [stringLenBytes / 2]byte
 
 	pos := 1
-	copy(b[:], s[pos:pos+uint8fieldBytes])
+	copy(b[:], t.String()[pos:pos+uint8fieldBytes])
```
YuriyNasretdinov, line 150: Not sure if you even need to call String() here to be honest because the underlying type is already string

src/meta/typeutils.go
```diff
@@ -120,204 +120,205 @@ func wrap(typ byte, byteFields []uint8, args ...string) string {
...
 		if r := recover(); r != nil {
-			res = fmt.Sprintf("panic!(orig='%s', hex='%x')", s, s)
+			res = fmt.Sprintf("panic!(orig='%s', hex='%x')", t, t)
 		}
 	}()
 
-	switch s[0] {
+	switch t[0] {
 	case WGlobal:
-		return "global_$" + formatType(UnwrapGlobal(s))
```
YuriyNasretdinov, line 290: Why did you remove formatting here? It is not present in the new `UnwrapGlobal` method

YuriyNasretdinov (changes requested):
More comments :)

src/meta/typesmap.go
```diff
@@ -36,11 +36,15 @@ func (m RawTypesMap) Add(typ Type) {
 	m[typ] = struct{}{}
 }
 
+func (m RawTypesMap) Delete(typ Type) {
```
YuriyNasretdinov, line 39: It does not make too much sense to me to declare methods like "Delete" and "Add" to value types

i582: Thanks for the comments!
Perhaps here Delete is too much, but `res.Add("some")` looks simpler than `res["some"] = struct{}{}`, that's why I introduced it.
Delete is just for a similar interface.

src/meta/typesmap.go
i582: Initially, this is just a copy of the `AppendString` function modified for types, in `AppendString` we consider that the map is mutable and imprecise, so this has lagged behind in this function as well.

src/meta/typeutils.go
i582: Now there is, thanks.

src/langsrv/refsvisitors.go
i582: I'll take a closer look. 
However, for `FindMethod`, this does not seem like a good idea, since the method takes the name of the class, not its type. Elsewhere we will have to cast the `string` to `Type`.

src/linter/block.go
i582: Initially, I wanted to avoid low-level interactions, such as here:

```go
	left.Iterate(func(typ string) {
		if typ[0] == meta.WArrayOf {
			hasAtLeastOneArray = true
		}
		if typ[0] == '\\' {
			hasAtLeastOneClass = true
		}
		newMap.AppendString(typ)
	})

	right.Iterate(func(typ string) {
		if typ[0] == meta.WArrayOf && meta.UnwrapArrayOf(typ) == "mixed" && hasAtLeastOneArray {
			return
		}
		if typ == "object" && hasAtLeastOneClass {
			return
		}
		newMap.AppendString(typ)
	})
```

So that we can conveniently work with individual types without going into implementation details and see all the methods that are available for them, and not a certain set of functions, such as `IsArrayType`, which for some reason is in the` meta` package.

i582:
Closed until better times.

Issue #866, opened 2020-12-21
Improvements in IR


i582:
* [ ] Reduce the number of fields with the `Node` type;
* [ ] Storing a flag for `ir.Assign` when assign represents an expression rather than a statement;
* [ ] For each field with the `Node[]` type, add a comment describing the possible types inside `Node`.

PR #867, opened 2020-12-21
ir: clarified types for some nodes

i582:
```go
ir.GroupUseStmt.UseType -> *ir.Identifier
ir.UseStmt.Use          -> *ir.Name
ir.UseListStmt.UseType  -> *ir.Identifier
```
#866

### DIFF
--- a/src/ir/node_types.go
+++ b/src/ir/node_types.go
@@ -531,7 +531,7 @@ type InstanceOfExpr struct {
 	FreeFloating freefloating.Collection
 	Position     *position.Position
 	Expr         Node
-	Class        Node
+	Class        Node // *ir.Identifier (static), *ir.Name (ClassNames, self), *ir.SimpleVar ($this)
 }
 
 // IssetExpr is a `isset($Variables...)` expression.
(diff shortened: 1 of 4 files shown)


src/ir/node_types.go
```diff
@@ -531,7 +531,7 @@ type InstanceOfExpr struct {
 	FreeFloating freefloating.Collection
 	Position     *position.Position
 	Expr         Node
-	Class        Node
+	Class        Node // *ir.Identifier (static), *ir.Name (ClassNames, self), *ir.SimpleVar ($this)
```
quasilyte, line 534: I think `Class` operand of the instance of can be literally anything.
A function call, index expression, etc.

src/ir/node_types.go
```diff
@@ -556,7 +556,7 @@ type MethodCallExpr struct {
 	FreeFloating freefloating.Collection
 	Position     *position.Position
 	Variable     Node
-	Method       Node
+	Method       Node // *ir.SimpleVar, *ir.Identifier
```
quasilyte, line 559: What about things like this?
```

class Foo {
    public function f() { var_dump('f()'); }
}

$foo = new Foo();
$foo->{'f'}();
```

quasilyte (approved):

PR #868, opened 2020-12-22
linter: improved closure and shape handling

### LINKED ISSUE #860: Error with closure
## Code Example

```php
<?php
class Foo {
  public function method() {}
}

function f1() {
  $f = function(): Foo { return new Foo; };
  $foo = f();
  $foo->method();
}
```

## Actual Behavior
```
<critical> ERROR   undefined: Call to undefined method {mixed}->method() at C:/projects/noverify/example/error.php:9
  $foo->method();
        ^^^^^^
```

## Expected Behavior

No error.

i582:
## Shapes
For shape, names are now generated using the keys that make up this shape.
This avoids duplication of classes for views of the same shape. 
The cache size has decreased by about 1.5%.

Before:
```php
\shape$filename.php$count_shapes$
```

Now:
```php
\shape$x:int,y:string$
```

## Closures

Now each closure is a function with a name that includes a file, a function,
and the number of line in the source code where it is declared.

For example:
```php
\Closure$(exprtype.php,func):7$
```

The functions for generating names have been moved to a separate package `autogen`.

Fixes #860


src/linter/block_linter.go
```diff
@@ -720,11 +720,17 @@ func (b *blockLinter) checkFunctionCall(e *ir.FunctionCallExpr) {
 	call := resolveFunctionCall(b.walker.ctx.sc, b.classParseState(), b.walker.ctx.customTypes, e)
 	fqName := call.funcName
 
-	if call.canAnalyze {
-		b.checkCallArgs(e.Function, e.Args, call.info)
-		b.checkDeprecatedFunctionCall(e, &call)
+	if !call.isClosure {
```
quasilyte, line 723: Is it enough to only check fo `!isClosure` here? It checked for canAnalyze previously as well.
Do we actually need canAnalyze now?

src/solver/closure.go
```diff
@@ -19,8 +23,50 @@ func IsSupportedFunction(name string) bool {
...
+	if !ok {
+		return meta.FuncInfo{}, false
+	}
+
+	var fi meta.FuncInfo
+	var found bool
+	var sv = &ir.SimpleVar{Name: nmf.Name}
+
+	tp := ExprTypeLocal(sc, cs, sv)
+	tp.Iterate(func(typ string) {
```
quasilyte, line 56: Maybe you want to use `Find` instead of `Iterate` here?

src/solver/closure.go
```diff
@@ -19,8 +23,50 @@ func IsSupportedFunction(name string) bool {
...
+	pos := ir.GetPosition(fun)
+	if curFunction != "" {
+		curFunction = "," + curFunction
+	}
+	name := `\Closure(` + curFile + curFunction + "):" + fmt.Sprint(pos.StartLine)
+	return name
+}
+
+func GetClosure(name ir.Node, sc *meta.Scope, cs *meta.ClassParseState) (meta.FuncInfo, bool) {
+	if !cs.Info.IsIndexingComplete() {
```
quasilyte, line 42: I don't think `GetClosure` should be called during indexing at all.
All other `Get*` methods expect to be called during linting only.
If they're called during the indexing phase, it's probably a mistake.

src/linter/block_linter.go
i582: This is ok, for checks from this body we do not need `canAnalyze`.

src/solver/closure.go
```diff
@@ -19,8 +23,50 @@ func IsSupportedFunction(name string) bool {
...
 }
 
-// Struct containing information about the function that called the closure.
+// ClosureCallerInfo containing information about the function that called the closure.
 type ClosureCallerInfo struct {
 	Name     string          // caller function name
 	ArgTypes []meta.TypesMap // types for each arg for call caller function
 }
+
+func GetClosureName(fun *ir.ClosureExpr, curFunction, curFile string) string {
```
quasilyte, line 32: Can we make it consistent with shape type names generation?
Right now shape-related code is located in `linter` package (generateShapeName).
Maybe we can put closure-related code there as well and then move them both if we understand if there is anything common between them? We'll also generate anonymous classes at some point, so I don't want autogen type related code be spread around a lot.

quasilyte:
Ping. Review comments are not addressed.

src/solver/closure.go
i582: Yep. Done.

quasilyte (approved):

PR #869, opened 2020-12-22
linter: dead code check extended

i582:
Added check for null coalesce when the type of the
the first operand is precise and does not contain null.


quasilyte (approved):

src/linter/block_linter.go
```diff
@@ -191,6 +192,17 @@ func (b *blockLinter) report(n ir.Node, level int, checkName, msg string, args .
 	b.walker.r.Report(n, level, checkName, msg, args...)
 }
 
+func (b *blockLinter) checkCoalesceExpr(n *ir.CoalesceExpr) {
+	lt := solver.ExprType(b.walker.ctx.sc, b.walker.r.ctx.st, n.Left)
```
quasilyte, line 196: What does `lt` mean though?

quasilyte: Left type?
I think `lhsType` is better here (or `leftType`).
`lt` is too alien and is not used anywhere in a project.

quasilyte (approved):

PR #870, opened 2020-12-23
embeddedrules: added precedence rules for ??

i582:
```
$_ [==, !=, ===, !==, <, >, <=, >=] $_ ?? $_
```


src/cmd/embeddedrules/rules.php
```diff
@@ -92,6 +92,46 @@ function precedence() {
     $_ !== $_ | $_;
     $_ | $_ !== $_;
   }
+
+  /** @warning === has higher precedence than ?? */
+  any_eq3_coalesce: {
```
quasilyte, line 97: I think we don't need `any_*` groups of 1 pattern.
Or there is some reason behind it?

src/tests/golden/testdata/embeddedrules/precedence.php
```diff
@@ -54,4 +63,13 @@ function precedence_lhs_good($x, $mask) {
...
   $_ = (precedence_foo() !== 0) & 0x02;
+
+  $_ = ($mask > $z) ?? $x;
+  $_ = ($mask < $z) ?? $x;
+  $_ = ($mask <= $z) ?? $x;
+  $_ = ($mask >= $z) ?? $x;
+  $_ = ($mask == $z) ?? $x;
+  $_ = ($mask != $z) ?? $x;
+  $_ = ($mask === $z) ?? $x;
+  $_ = ($mask !== $z) ?? $x;
```
quasilyte, line 74: Please add forms like `$x ?? $mask > $z` so we don't screw this up later as these forms are valid.

quasilyte (approved):

PR #873, opened 2021-01-11
cmd: associate absolute paths with git files

i582:
When running NoVerify in normal (non-git) mode,
all filenames are turned into full filenames (abs paths).
It's not the case with git mode, since most code paths
recorded relative names.

Evaluate .git-containing folder once, during linter init
and use it to get absolute path later, when collecting
filenames provided by git catter.

### DIFF
--- a/src/cmd/check.go
+++ b/src/cmd/check.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"flag"
 	"fmt"
+	"log"
+	"path/filepath"
 
 	"github.com/VKCOM/noverify/src/linter"
 )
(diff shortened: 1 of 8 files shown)


src/cmd/check.go
```diff
@@ -26,6 +28,13 @@ func Check(cfg *MainConfig) (int, error) {
 	if args.disableCache {
 		linter.CacheDir = ""
 	}
+	if args.gitRepo != "" {
+		gitDir, err := filepath.Abs(args.gitRepo + "/../")
+		if err != nil {
+			log.Fatalf("Find git dir: %v", err)
```
quasilyte, line 34: Don't do log fatal. You can return an error here.

quasilyte (approved):

PR #875, opened 2021-01-12
linter: some fields from RootWalker have been moved to the workspace.File structure

i582:
Thanks to this change, the number of fields in the
`RootWalker` has decreased, and the signatures of some
functions have been simplified, which now accept a
single argument of type `*workspace.File`.


src/workspace/file.go
```diff
@@ -0,0 +1,151 @@
...
+	"strings"
+
+	"github.com/VKCOM/noverify/src/git"
+)
+
+type File struct {
+	name     string
+	contents []byte
+
+	lineRanges []git.LineRange
```
quasilyte, line 14: Let's not store git-related info in that?

src/workspace/file.go
```diff
@@ -0,0 +1,151 @@
...
+		pos += len(ln) + 1
+	}
+	return lines, linesPos
+}
+
+func (f *File) Name() string {
+	return f.name
+}
+
+func (f *File) CountLines() int {
```
quasilyte, line 54: I would prefer `NumLines()` name to be consistent with `go/types` package.

src/linter/worker.go
```diff
@@ -106,9 +105,11 @@ func (w *Worker) ParseContents(filename string, contents []byte, lineRanges []gi
 	b := w.ctx.scratchBuf
 	b.Reset()
 	b.ReadFrom(rd)
-	contents = append(make([]byte, 0, b.Len()), b.Bytes()...)
 
-	waiter := BeforeParse(len(contents), filename)
+	contents := append(make([]byte, 0, b.Len()), b.Bytes()...)
+	file.ResetContents(contents)
```
quasilyte, line 110: Is it possible to create a new `File` here instead of changing existing one?

I also don't think that it's a win to replace FileInfo with File during the filesystem traversal.
We sometimes don't have a file "contents", so our `File` can be in two states: with data and without date.
It would be simpler if `File` is guaranteed to always be a file with contents.
It's an entity used for linter, while FileInfo is mostly for the traversal-related code and to create`File` later.

src/workspace/file.go
```diff
@@ -0,0 +1,151 @@
...
+	f.contents = newContents
+	f.lines, f.linesPos = parseContents(f.contents)
+	f.autoGenerated = FileIsAutoGenerated(f.contents)
+}
+
+func (f *File) Contents() []byte {
+	return f.contents
+}
+
+func (f *File) Part(start, end int) []byte {
```
quasilyte, line 95: I'm not 100% sure that this method is needed. Most places should just slice `f.Contents` directly and 1 place should do the index checking itself. Making `File` as a mostly-data object is preferable. If you want some fancy accessing thing, you can write a helper function on the spot. Attaching a method that is only useful in 1 place makes API bigger than we need without a good reason (other places index []byte slice without any checks).

src/linter/root.go
```diff
@@ -111,20 +98,9 @@ func NewWalkerForReferencesSearcher(workerCtx *WorkerContext, filename string, b
...
-		linesPositions[idx] = pos
-		pos += len(ln) + 1
-	}
-
-	d.fileContents = contents
-	d.LinesPositions = linesPositions
-	d.Lines = lines
-	d.autoGenerated = workspace.FileIsAutoGenerated(contents)
+// InitFileData initializes file that are needed for RootWalker work
+func (d *RootWalker) InitFileData(filename string, contents []byte, lineRanges []git.LineRange) {
```
quasilyte, line 102: I don't think this method needs `lineRanges` parameter?

quasilyte (approved):

PR #884, opened 2021-01-15
remove langsrv

quasilyte:
See https://github.com/VKCOM/noverify/releases/tag/v0.3.0 description

We remove language server support until the refactoring is complete.

There are several steps that are required, but are difficult with
language server that is too coupled with linter package internals.
It also goes into the global state.

We may re-implement language server when the time is right.


quasilyte:
Press `F` to Pay Respects.

i582:
The whole NoVerify team:

PR #893, opened 2021-01-15
linter: reduced the number of report levels

i582:
`LevelSyntax` -> `LevelError`
`LevelMaybe`, `LevelUnused`, `LevelHint` -> `LevelNotice`

`LevelInformation` renamed to `LevelInfo`.


src/linter/block.go
```diff
@@ -495,7 +495,7 @@ func (b *blockWalker) walkComments(n ir.Node, c freefloating.String) {
 
 		types, warning := typesFromPHPDoc(&b.r.ctx, p.Type)
 		if warning != "" {
-			b.r.Report(n, LevelInformation, "phpdocType", "%s on line %d", warning, p.Line())
+			b.r.Report(n, LevelWarning, "phpdocType", "%s on line %d", warning, p.Line())
```
quasilyte, line 498: I think phpdoc mistake is a notice, not a warning. It's not even a part of the executable PHP code.

src/linter/block_linter.go
```diff
@@ -297,14 +297,14 @@ func (b *blockLinter) checkBinaryDupArgs(n, left, right ir.Node) {
 // checkVoidType reports if node has void type
 func (b *blockLinter) checkVoidType(n ir.Node) {
 	if b.walker.exprType(n).Is("void") {
-		b.report(n, LevelDoNotReject, "voidResultUsed", "void function result used")
+		b.report(n, LevelNotice, "voidResultUsed", "void function result used")
```
quasilyte, line 300: This is usually a serious logical error and it's easily fixed with a discard var assignment.
I would mark it as error as it will cause a program to misbehave.

src/linter/root.go
```diff
@@ -1053,7 +1047,7 @@ func (d *rootWalker) parsePHPDocVar(n ir.Node, doc []phpdoc.CommentPart) (m meta
 		if ok && part.Name() == "var" {
 			types, warning := typesFromPHPDoc(&d.ctx, part.Type)
 			if warning != "" {
-				d.Report(n, LevelInformation, "phpdocType", "%s on line %d", warning, part.Line())
+				d.Report(n, LevelWarning, "phpdocType", "%s on line %d", warning, part.Line())
```
quasilyte, line 1050: Same as for other phpdocType. It should probably be a notice.

quasilyte:
Seems like some tests were not updated.

PR #895, opened 2021-01-15
WIP: all: make linter config explicit (instead of being global vars)

quasilyte:
Since global usages were chaotic, this change affected a lot of files.


src/linter/conf.go
```diff
@@ -33,28 +33,36 @@ var (
 	// SrcInput implements source code reading from files and buffers.
 	//
 	// TODO(quasilyte): avoid having it as a global variable?
```
i582, line 35: Probably, the comment about globality can be removed.

quasilyte: Fixed.

PR #898, opened 2021-01-17
linter: move PHPExtensions to linter.Config

i582:


src/linter/linter.go
```diff
@@ -92,7 +91,7 @@ func (l *Linter) analyzeFiles(readFileNamesFunc workspace.ReadCallback, allowDis
 
 func (l *Linter) InitStubs(readFileNamesFunc workspace.ReadCallback) {
 	meta.SetLoadingStubs(true)
-	l.analyzeFiles(readFileNamesFunc, nil)
```
quasilyte, line 95: I'm not sure this is needed.
When Stubs are initialized, we don't need to apply any kind of filters.
Is there a reason behind this line change? It's not related to PHP extentions.

src/linter/linter.go
```diff
@@ -34,10 +33,10 @@ func (l *Linter) NewIndexingWorker(id int) *Worker {
 
 // AnalyzeFiles runs linter on the files that are provided by the readFileNamesFunc function.
 func (l *Linter) AnalyzeFiles(readFileNamesFunc workspace.ReadCallback) []*Report {
-	return l.analyzeFiles(readFileNamesFunc, l.config.AllowDisable)
+	return l.analyzeFiles(readFileNamesFunc)
 }
 
-func (l *Linter) analyzeFiles(readFileNamesFunc workspace.ReadCallback, allowDisable *regexp.Regexp) []*Report {
+func (l *Linter) analyzeFiles(readFileNamesFunc workspace.ReadCallback) []*Report {
```
quasilyte, line 39: See another comment. I'm not sure this method should be changed.
At least it's not right to change it as a part of PHP extensions related PR.

src/cmd/check.go
```diff
@@ -27,9 +28,12 @@ func Check(cfg *MainConfig) (int, error) {
 	var args cmdlineArguments
 	bindFlags(config, ruleSets, &args)
 	flag.Parse()
+
 	if args.disableCache {
 		config.CacheDir = ""
 	}
+	config.PhpExtensions = strings.Split(args.phpExtensionsArg, ",")
```
quasilyte, line 35: This is not the best place to initialize the config.
linterRunner does the args->config assignments.

quasilyte (approved):

PR #902, opened 2021-01-20
ir: added tokens to ir nodes

i582:
This is the first step in migrating to the new
a version of the php parser.
The new version of the parser moved away from
storing everything in free-floating and switched
to the concept of tokens.

### DIFF
--- a/src/ir/node_types.go
+++ b/src/ir/node_types.go
@@ -6,6 +6,10 @@ import (
 	"github.com/VKCOM/noverify/src/phpdoc"
 )
 
+// Token is a stub while switching to a new version of the parser.
+// Replace later to token.Token from php-parser
+type Token struct{}
+
 // TODO: make Alt and AltSyntax field names consistent.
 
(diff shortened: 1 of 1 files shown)


src/ir/node_types.go
```diff
@@ -662,37 +802,44 @@ type TernaryExpr struct {
 	FreeFloating freefloating.Collection
 	Position     *position.Position
 	Condition    Node
+	QuestionTkn  *Token
 	IfTrue       Node
+	ColonTkn     *Token
 	IfFalse      Node
 }
```
quasilyte, line 809: What about grouping tokens separately from the other fields?
Like this:
```
type TernaryExpr struct {
	FreeFloating freefloating.Collection
	Position     *position.Position
	Condition    Node
	IfTrue       Node
	IfFalse      Node

	ColonTkn     *Token
	QuestionTkn  *Token
}
```

quasilyte (approved):

PR #904, opened 2021-01-20
ir: add Node.IterateTokens method

i582:
Since we are moving from free-floating to tokens,
we need to be able to iterate all tokens regardless
of the node type.


src/ir/codegen/gen_iterate_tokens.go
```diff
@@ -0,0 +1,48 @@
...
+	})
+}
+
+func (g *genIterate) writeIterate(w *bytes.Buffer, pkg *packageData, typ *typeData) {
+	for i := 0; i < typ.info.NumFields(); i++ {
+		field := typ.info.Field(i)
+		switch typeString := field.Type().String(); typeString {
+		case "*ir.Token": // replace later with *github.com/z7zmey/php-parser/pkg/token.Token
+			fmt.Fprintf(w, "    if n.%[1]s != nil {\n", field.Name())
+			fmt.Fprintf(w, "        if !cb(n.%[1]s) {\n", field.Name())
```
quasilyte, line 36: When fmt arg is only used once, `%s` is preferred to `%[1]s`.

src/ir/codegen/gen_iterate_tokens.go
```diff
@@ -0,0 +1,48 @@
...
+			fmt.Fprintf(w, "            return\n")
+			fmt.Fprintf(w, "        }\n")
+			fmt.Fprintf(w, "    }\n")
+		case "[]*ir.Token": // replace later with []*github.com/z7zmey/php-parser/pkg/token.Token
```
quasilyte, line 40: Add a `TODO:` prefix to your comments so they can be noticed later.
It's convenient to grep for TODO.

src/ir/codegen/gen_iterate_tokens.go
i582: Done.

quasilyte (approved):

PR #906, opened 2021-01-21
ir: added CurlyBrace flag to ArrayDimFetchExpr, which indicates whether the brace is a curly brace

i582:


src/ir/irconv/irconv_test.go
```diff
@@ -313,3 +318,59 @@ func TestInterpretString(t *testing.T) {
...
+			curlyBrace: true,
+		},
+		{
+			code:       `$arr['x'   ];`,
+			curlyBrace: false,
+		},
+	}
+
+	for _, test := range tests {
+		p := php7.NewParser([]byte(`<?php ` + test.code))
```
quasilyte, line 362: Is it possible to use parseutil function for parsing here?
It's painful to use the parser directly in different places as we're going to migrate soon.

quasilyte (changes requested):
Please add some phpgrep tests and update the matcher code so it handles this flag correctly.

i582:
> Please add some phpgrep tests and update the matcher code so it handles this flag correctly.

Maybe in the next PR?

i582:
There are not many changes, so let it be in this PR.

src/phpgrep/matcher_test.go
```diff
@@ -422,6 +422,8 @@ func TestMatch(t *testing.T) {
 		{`$x->ff(1, 2)`, `$this->ff(1, 2)`},
 
 		{`$_[0]`, `$v[0]`},
+		{`$_{0}`, `$v{0}`},
+		{`$_{0}`, `$v[0]{0}`},
```
quasilyte, line 426: You also need to add negative tests.
For example, `$_[$_]` should not match `$_{$_}` and vice versa.

PR #907, opened 2021-01-21
linter: added a check for the use of curly braces for taking an element by index

i582:


src/linter/block_linter.go
```diff
@@ -569,6 +572,12 @@ func (b *blockLinter) checkContinueStmt(c *ir.ContinueStmt) {
 	}
 }
 
+func (b *blockLinter) checkArrayDimFetch(s *ir.ArrayDimFetchExpr) {
+	if s.CurlyBrace {
+		b.report(s, LevelDoNotReject, "arraySyntax", "Don't use curly braces to take an item by index (use square braces form instead)")
```
quasilyte, line 577: Maybe this text would be easier to understand?
```
a{i} indexing is deprecated since PHP 7.4, use a[i] instead
```

quasilyte:
I also think that it should be a separate checker.
It has nothing to deal with `arraySyntax`, which is about array literals.
(Not to mention that this checker also handles strings and ArrayAccess classes.)

I would try making it an embedded rule with `@fix` that changes the indexing notation.
```
$a{$i} -> $a[$i]
```

quasilyte (changes requested):
It should not be `arraySyntax`.

quasilyte (approved):

PR #912, opened 2021-01-26
linter: fixed processing of variables in try-catch

i582:
Now, if a variable is declared in `try` and all `catch`, or `finally`, 
then it will not give a warning that the variable may not be defined.


src/linter/block.go
```diff
@@ -582,7 +582,9 @@ func (b *blockWalker) withNewContext(action func()) *blockContext {
 }
 
 func (b *blockWalker) handleTry(s *ir.TryStmt) bool {
-	contexts := make([]*blockContext, 0, len(s.Catches)+1)
+	var contexts = make([]*blockContext, 0, len(s.Catches)+1)
```
quasilyte, line 585: Why change `:=` to var?

src/tests/checkers/catch_test.go
```diff
@@ -110,3 +110,111 @@ try {
...
+	test := linttest.NewSuite(t)
+	test.AddFile(`<?php
+function f() {
+	try {
+		$a = 100;
+		$c = 100;
+		$d = 100;
+	} catch (Exception $_) {
+		$a = 200;
+	} catch (ExceptionDerived $_) {
```
quasilyte, line 124: It seems like we don't give "undefined class" warnings for catch clauses?
This test case should probably define `ExceptionDerived` or do a RunFilterMatch so it's less prone to unrelated changes.

src/linter/block.go
i582: Just so that all variable definitions are the same :)
```go
var contexts = make([]*blockContext, 0, len(s.Catches)+1)
var linksCount int
var finallyCtx *blockContext
```

src/tests/checkers/catch_test.go
i582: > It seems like we don't give "undefined class" warnings for catch clauses?

It seems so.

src/linter/block.go
quasilyte: `var` is generally used for "zero init".
Please don't use var with assignments anywhere except global scope.

quasilyte (approved):

PR #914, opened 2021-01-27
linter: added check and quickfix for the presence of a comma for the last element in a multi-line array

### LINKED ISSUE #833: Comma afet last array element
## Code Example

```php
$array = [
    1,
    2,
    3
];
```

## Actual Behavior

Nothing

## Expected Behavior

I would expect to see some error to add comma after last element.

Thanks!

i582:
Fixes #833


src/linter/block_linter.go
```diff
@@ -675,6 +679,38 @@ func (b *blockLinter) checkArray(arr *ir.ArrayExpr) {
...
+		return
+	}
+
+	item := items[currentItemIndex]
+
+	from := item.Position.StartPos
+	to := item.Position.EndPos
+	itemText := b.walker.r.file.Contents()[from : to+1]
+	if itemText[len(itemText)-1] != ',' && itemText[len(itemText)-1] != ']' {
+		b.report(item, LevelWarning, "trailingComma", "last element in a multi-line array must have a trailing comma")
```
quasilyte, line 709: Maybe it's a notice? It's a minor style issue.

src/linter/block_linter.go
```diff
@@ -598,11 +598,15 @@ func (b *blockLinter) checkArray(arr *ir.ArrayExpr) {
...
 
-	for _, item := range items {
+	multiline := arr.Position.EndLine-arr.Position.StartLine != 0
+
+	for index, item := range items {
 		if item.Val == nil {
 			continue
 		}
 
+		b.checkMultilineArrayTrailingComma(multiline, index, items)
```
quasilyte, line 608: I don't exactly like passing a bool and then return inside a function based on that argument.

It's better to check before calling a function or to pass `arr` there and check for the multiline inside it.
`multiline` variable is not needed inside checkArray function anyway.

src/linter/block_linter.go
```diff
@@ -598,11 +598,15 @@ func (b *blockLinter) checkArray(arr *ir.ArrayExpr) {
 	haveImplicitKeys := false
 	keys := make(map[string]ir.Node, len(items))
 
-	for _, item := range items {
+	multiline := arr.Position.EndLine-arr.Position.StartLine != 0
```
YuriyNasretdinov, line 601: First of all, you could write that as
```go
multiline := arr.Position.EndLine != arr.Position.StartLine
```

Second of all, I would suggest to not try to be clever and just write it like this:

```go
multiline := false
if arr.Position.EndLine != arr.Position.StartLine {
multiline = true
}
```

The last version is much easier to understand.

src/linter/block_linter.go
```diff
@@ -675,6 +679,38 @@ func (b *blockLinter) checkArray(arr *ir.ArrayExpr) {
...
+func (b *blockLinter) checkMultilineArrayTrailingComma(multiline bool, currentItemIndex int, items []*ir.ArrayItemExpr) {
+	if !multiline || currentItemIndex != len(items)-1 {
+		return
+	}
+
+	item := items[currentItemIndex]
+
+	from := item.Position.StartPos
+	to := item.Position.EndPos
+	itemText := b.walker.r.file.Contents()[from : to+1]
```
YuriyNasretdinov, line 707: You don't check that there is enough length in the contents. The code might panic.

YuriyNasretdinov (approved):

src/linter/block_linter.go
i582: Done

src/linter/block_linter.go

src/linter/block_linter.go

quasilyte (approved):
I wonder though: should it be enabled by default?
@YuriyNasretdinov what do you think?
It may be very pedantic for some users.

quasilyte (approved):

quasilyte (approved):

PR #916, opened 2021-01-27
cmd: small refactor rule init code

i582:
The `Config.Rules` field is now fully initialized, including
the fields of the `rules.Set` structure. Thus, we got rid of the
need to initialize this field every time.

The `InitEmbeddedRules` function has been renamed to `AddEmbeddedRules`
and now accepts not the config, but the rules directly, as well
as for the `appendRuleSet` function.

### DIFF
--- a/src/cmd/rules.go
+++ b/src/cmd/rules.go
@@ -11,17 +11,17 @@ import (
 	"github.com/VKCOM/noverify/src/rules"
 )
 
-func AddEmbeddedRules(rulesSet *rules.Set, p *rules.Parser, filter func(r rules.Rule) bool) ([]*rules.Set, error) {
-	ruleSets, err := parseEmbeddedRules(p)
+func AddEmbeddedRules(rset *rules.Set, p *rules.Parser, filter func(r rules.Rule) bool) ([]*rules.Set, error) {
+	embeddedRules, err := parseEmbeddedRules(p)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, rset := range ruleSets {
-		appendRuleSet(rulesSet, rset, filter)
+	for _, embeddedRule := range embeddedRules {
+		appendRuleSet(rset, embeddedRule, filter)
 	}
 
-	return ruleSets, nil
+	return embeddedRules, nil
 }
 
 func parseRules() ([]*rules.Set, error) {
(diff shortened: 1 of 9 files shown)


src/cmd/rules.go
```diff
@@ -8,19 +8,19 @@ import (
 	"strings"
 
 	"github.com/VKCOM/noverify/src/cmd/embeddedrules"
-	"github.com/VKCOM/noverify/src/linter"
 	"github.com/VKCOM/noverify/src/rules"
 )
 
-func InitEmbeddedRules(config *linter.Config, p *rules.Parser, filter func(r rules.Rule) bool) ([]*rules.Set, error) {
+func AddEmbeddedRules(rules *rules.Set, p *rules.Parser, filter func(r rules.Rule) bool) ([]*rules.Set, error) {
```
quasilyte: You can call this param `rset`.

i582: `rset` is already used inside the loop

src/linter/conf.go
```diff
@@ -68,7 +68,7 @@ func NewConfig() *Config {
 	addBuiltinCheckers(reg)
 	return &Config{
 		SrcInput:       inputs.NewDefaultSourceInput(),
-		Rules:          &rules.Set{},
+		Rules:          rules.NewSet(),
```
quasilyte, line 71: There is a problem here.
cloneRulesForFile from linter/utils.go will now copy rules sets even if rules are not used.
We need to change `ruleSet != nil` check with something else as they're never nil now.

src/cmd/rules.go
```diff
@@ -8,20 +8,21 @@ import (
...
-	ruleSets, err := parseEmbeddedRules(p)
+func AddEmbeddedRules(rset *rules.Set, p *rules.Parser, filter func(r rules.Rule) bool) ([]*rules.Set, error) {
+	embeddedRules, err := parseEmbeddedRules(p)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, rset := range ruleSets {
-		appendRuleSet(config, rset, filter)
+	for _, embeddedRule := range embeddedRules {
```
quasilyte, line 21: It's not a rule, but a rule set.
Maybe remove the embeddedRule variable and use `embeddedRules[i]` to pass it as an argument?

i582: Can be renamed to `embeddedRuleSet` and `embeddedRules` to `embeddedRuleSets`.

quasilyte (approved):

PR #918, opened 2021-01-29
rules: added check for the presence of a variable in the template

### LINKED ISSUE #754: Give a rule compile error if filters refer to the variables that are not present in the pattern
If pattern does not refer to `$x` we should probably give a rule compilation error:
```php
/**
 * @type int $x
 */
$_ = foo();
```

i582:
Fixes #754


src/ir/irutil/irtutil.go
```diff
@@ -66,6 +66,29 @@ func FmtNode(n ir.Node) string {
 	return irfmt.Node(n)
 }
 
+// Find searches for a node in the passed subtree.
+func Find(what ir.Node, where ir.Node) bool {
```
quasilyte, line 70: Is this functions used?

i582: No, left for the future.

i582: By the way, we have already discussed that it is possible 
to collect information about variables and check them here, 
but I don’t remember exactly, but there were some problems 
that I could not solve.

quasilyte: That's not a good reason to keep unused code. :)

quasilyte: These util functions are almost never removed, even if unused. It's a good way to keep sanity with not adding more if you don't need them that much.

i582: Removed.

PR #925, opened 2021-02-08
all: switch to the new version of the parser

i582:
# Switching to the new version of the parser

- The `freefloating` has been completely removed.

- Due to the fact that the new version of the parser uses 
the concept of `tokens`, rather than `freefloating`, a 
large number of changes are associated with handling comments.

- Most of the changes are in `irconv.go`. Due to the fact that some nodes 
do not exist in the parser, the transformation logic has been added.

- To store `phpdoc` in nodes, a new type `ir.Doc` has been created.

- Keyword validation is now token-based. The `irutil.Keywords` function 
has been created to obtain the required tokens for a node.

- The `ir.CaseListStmt` node was removed as redundant.

- `IterateTokens` now traverses `FreeFloatings` inside the token.

- Removed the entire vendor parser.

- Added `GetFirstToken` function that returns the first non-nil node token.
   This is necessary to implement style checking since in the new version of the parser, all spaces and line breaks are attached to the first existing token (including if a node has a different node first, then the necessary information will be in the first token of this node).

- Full transfer to the parser upstream.


src/ir/codegen/gen_iterate_tokens.go
```diff
@@ -13,16 +13,38 @@ func (g *genIterate) Run() error {
 	ctx := g.ctx
 
 	var buf bytes.Buffer
+
+	fmt.Fprintf(&buf, `func handleToken(t *token.Token, cb func(*token.Token) bool) bool {
```
quasilyte, line 17: Can this function be defined outside of the auto-generated code?
There are examples of this, when normal functions call auto-generated functions.
It's not very convenient to write such code inside strings; and I think there are no strong reasons for this here.

quasilyte: This function also needs a comment.
It's not 100% obvious what bool result means.
In particular, it returns true for nil tokens, so it's not "if at least 1 callback returned true".
The behavior should be either simple to get right, or well-docummented.
I'll take a look at its usages, but right now this exact behavior looks odd.

src/ir/codegen/gen_iterate_tokens.go
```diff
@@ -13,16 +13,38 @@ func (g *genIterate) Run() error {
...
+		return true
+	}
+	
+	if !cb(t) {
+		return false
+	}
+
+	needReturn := true
+	for _, ff := range t.FreeFloating {
+		needReturn = needReturn && handleToken(ff, cb)
```
quasilyte, line 28: I'm not a fan of such loops as:
1. When `needReturn` is false, iteration will continue, but `handleToken` will not be called, maybe we can return without it?
2. It's not obvious what is the intended behavior. I can't distinguish it from a bug.

src/ir/codegen/gen_iterate_tokens.go
i582: Done.

src/ir/codegen/gen_iterate_tokens.go
i582: Done.

quasilyte (approved):
Fingers crossed. :crossed_fingers:

PR #939, opened 2021-03-01
linttest: add inline tests

i582:
New test method similar to [analysistest](https://pkg.go.dev/golang.org/x/tools/go/analysis/analysistest).

Expected errors are written on the same line in the comment as the expected error.

For example:

```php
<?php

function f() {
    $x = 100;

    $y; // want `expression evaluated but not used` and `Undefined variable: y`
    $x; // want `expression evaluated but not used`

    $_ = new Foo; // want `Type \Foo not found`
}
```

The following template is used to describe expected errors:
```
want `error description 1` [and `error description 2` and `error 3` ...]
```

If the error was not caught, then the test will fail. 
If not all errors are caught, the test will fail. 
If an error was caught that does not exist, then the test will fail.


YuriyNasretdinov:
I like the idea.

quasilyte:
Note that analysistest uses regular expressions in `want` pattern.
I'm not a big fan of that as 95% of the time I use `\Q` regexp escape to write a plain text.

But for our tests, I think strict strict comparison is also quite harsh.
Maybe we should check for the substring? I.e. `strings.Contains(reported, expected)`

i582:
Done check with substring.

PR #942, opened 2021-03-03
cmd/embeddedrules: added a check for define, which prohibits the use …

i582:
…of the third argument for specifying case-insensitive constants


src/cmd/embeddedrules/rules.php
```diff
@@ -368,3 +368,21 @@ function indexingSyntax() {
      */
     $x{$y};
 }
+
+/**
+ * @comment Report a call define function with third arguments true or false to define case (in)sensitive constants.
+ * @before  define("Z_CONST", 1, true);
+ * @after   define("Z_CONST", 1);
+ */
+function defineConst() {
```
quasilyte, line 377:
```
defineConst -> caseInsensitiveDefine
```

src/cmd/embeddedrules/rules.php
```diff
@@ -368,3 +368,21 @@ function indexingSyntax() {
      */
     $x{$y};
 }
+
+/**
+ * @comment Report a call define function with third arguments true or false to define case (in)sensitive constants.
```
quasilyte, line 373:
```
Report a call define function with third arguments true or false to define case (in)sensitive constants.
```
=>
```
Report a define() call with 3 arguments.
```

src/cmd/embeddedrules/rules.php
quasilyte: Technically, it's also possible to give a case sensitive param. 🤔
Maybe `define3` then?
`defineConst` sounds too generic and it doesn't describe what's the problem.
The problem is 3rd argument.
We could also add this check in `langDeprecated` perhaps.

PR #943, opened 2021-03-03
cmd/embeddedrules: added check for int needle

i582:


quasilyte:
Could you add a quickfix test?

src/cmd/embeddedrules/rules.php
```diff
@@ -368,3 +368,26 @@ function indexingSyntax() {
      */
     $x{$y};
 }
+
+/**
+ * @comment Report using an integer for needle.
```
quasilyte, line 373:
```
Report using an integer for needle
```
=>
```
Report using an integer for $needle argument of str* functions
```

quasilyte:
Also, I would add tests with non-precise type arguments.

src/cmd/embeddedrules/rules.php
i582: Done.

i582:
> Also, I would add tests with imprecise type arguments.

Added.

i582:
> Could you add a quickfix test?

The presence of `@fix` in the rule seems to be a bug.
I don’t think it’s a good idea to add the same description for 
each function in the rule in order to add a fix that is most likely 
not completely safe.

quasilyte (approved):

PR #945, opened 2021-03-04
linter: added order check in nested ternary operator

i582:
See [deprecated core nested-ternary](https://www.php.net/manual/ru/migration74.deprecated.php#migration74.deprecated.core.nested-ternary).

### DIFF
--- a/src/linter/block_linter.go
+++ b/src/linter/block_linter.go
@@ -434,6 +434,25 @@ func (b *blockLinter) checkTernary(e *ir.TernaryExpr) {
 		return // Skip `$x ?: $y` expressions
 	}
 
+	// containsParenTernary checks if a node is a parenthesized ternary operator.
+	containsParenTernary := func(n ir.Node) (containsTernary, hasBracket bool) {
+		switch n := n.(type) {
+		case *ir.ParenExpr:
+			_, containsTernary = n.Expr.(*ir.TernaryExpr)
+			return containsTernary, true
+
+		case *ir.TernaryExpr:
+			return true, false
+		}
+
+		return false, false
+	}
+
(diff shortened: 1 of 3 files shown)


src/linter/report.go
```diff
@@ -628,6 +628,17 @@ echo $someVal;`,
 	20,
 ]`,
 		},
+
+		{
+			Name:     "ternaryOrder",
```
quasilyte, line 633: How about `nestedTernary`?

i582: Done.

src/linter/block_linter.go
```diff
@@ -434,6 +434,25 @@ func (b *blockLinter) checkTernary(e *ir.TernaryExpr) {
...
+			return true, false
+		}
+
+		return false, false
+	}
+
+	containsTernary, hasBracket := containsParenTernary(e.Condition)
+	if !hasBracket && containsTernary {
+		b.report(e.Condition, LevelWarning, "ternaryOrder", "in ternary operators, you must explicitly use parentheses to specify the order of operations")
+	}
```
quasilyte, line 454: I may be wrong, but `containsParenTernary` could probably be removed and code may be simplified to this:
```go
_, nestedTernary := e.Condition.(*ir.TernaryExpr)
if nestedTernary {
  b.report(e.Condition, LevelWarning, "ternaryOrder", "in ternary operators, you must explicitly use parentheses to specify the order of operations")
}
```
It seems like we're only interested in nested ternary; if there is a paren expr, we'll not match it against `*ir.TernaryExpr`, so that condition seems redundant.

i582: You're absolutely right! Excellent simplification.

quasilyte (approved):

PR #946, opened 2021-03-04
linter: added a check for casting to the real type and using the is_real function

i582:
See [deprecated core real](https://www.php.net/manual/ru/migration74.deprecated.php#migration74.deprecated.core.real).


src/linter/report.go
```diff
@@ -628,6 +628,17 @@ echo $someVal;`,
 	20,
 ]`,
 		},
+
+		{
+			Name:     "langDeprecated",
+			Default:  false,
+			Quickfix: true,
+			Comment:  `Report the use of deprecated (per language spec) symbols.`,
```
quasilyte, line 636: s/symbols/features/

src/tests/checkers/basic_test.go
```diff
@@ -2126,3 +2126,22 @@ function f() {
...
+function f() {
+    $a = (real)100;
+    if (is_real($a)) {
+        echo 1;
+    }
+}
+`)
+	test.Expect = []string{
+		`do not cast to type real, use cast to float`,
+		`don't use is_real function, use is_float`,
```
quasilyte, line 2144: I would stick to this style of suggestion:
```
use is_float function instead of is_real
```

i582: Done.

src/linter/report.go
i582: Done.

quasilyte (approved):

PR #950, opened 2021-03-04
linter: added a check for calling array_key_exists with an object

i582:


src/tests/checkers/basic_test.go
```diff
@@ -2174,3 +2174,40 @@ function f() {
...
+function returnObjectAndNull(): ?Foo {
+	if (1) {
+		return null;
+	}
+	return new Foo;
+}
+
+function f() {
+    $foo = new Foo;
+	$arr = ["a" => 100];
```
quasilyte, line 2198: Minor: this line looks to be aligned inconsistently with other lines.

i582: Fixed.

PR #951, opened 2021-03-09
cmd: added tracking of the number of reports for tests (potentially f…

i582:
…or a side project)


quasilyte (commented):
Why this PR also includes color-related dependencies?
It looks like a mistake, these libraries are not used here.

i582:
Since this will only work for VKCOM members, it doesn't make much sense.

PR #955, opened 2021-03-24
linter: added check of comments /* */ for the fact that they are in f…

### LINKED ISSUE #954: Report "not quite phpdoc" comments
## Code Example

```php
class Foo {
  /*
   * @return int
   */
  public function f() { return 1; }
}
```

Corrected code example:

```php
class Foo {
  /**
   * @return int
   */
  public function f() { return 1; }
}
```

## Actual Behavior

No warnings.

## Expected Behavior

Warning about a malformed phpdoc comment.

```
phpdocLint: phpdoc comment should start with /**, not /*
```

Note: there is 1 case where we want to allow it.

This is OK:
```php
/* @var int $foo */
$foo = f();
```

Most tools understand a single-line phpdoc in `/* @tag */` format.

How to detect:

* Comment starts with `/*` (not `/**`), but contains `@` annotations.

i582:
…act phpdoc, but when writing there was a mistake and only one asterisk was put at the beginning

Fixes #954


src/phpdoc/parser.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/z7zmey/php-parser/pkg/token"
 )
 
+type PhpDoc struct {
```
quasilyte, line 10: In Go, it's idiomatic to avoid stutter where possible.
Package name is a part of the type prefix, so you can call this struct `Comment`.
```diff
- phpdoc.PhpDoc
+ phpdoc.Comment
```

src/phpdoc/parser.go
```diff
@@ -58,22 +64,47 @@ func IsPHPDocToken(t *token.Token) bool {
...
+		return IsPossiblePhpDoc(t.Value)
 	}
 
 	return true
 }
 
+// IsPossiblePhpDoc checks if /* */ comments contain annotations, which may mean that
+// it is phpdoc, but there is a mistake when there is one asterisk instead of two at
+// the beginning of a comment.
+func IsPossiblePhpDoc(value []byte) bool {
```
quasilyte, line 76: Does phpstorm understand any kind of a single-line phpdoc with `/* @tag */` syntax?
If it does, we could check for any tag, not for a tag from a list.
And rename a function to `ContainsTag` (again, mentioning phpdoc in a name is not always needed if this function is a part of phpdoc package).

src/tests/checkers/phpdoc_test.go
```diff
@@ -425,3 +425,41 @@ func TestPHPDocIncorrectSyntaxOptionalTypesType(t *testing.T) {
 	}
 	test.RunAndMatch()
 }
+
+func TestPHPDocInvalidBeginning(t *testing.T) {
```
quasilyte, line 429: Could you add tests that make sure that a single line phpdocs (not only `@var`) do not cause any warnings?

src/phpdoc/parser.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/z7zmey/php-parser/pkg/token"
 )
 
+type PhpDoc struct {
+	Raw        string
+	Parsed     []CommentPart
+	Suspicious bool
```
quasilyte, line 13: As I see, we use this field in 2 unconnected places only once per comment.
Would it be hard to get rid of this field and check whether a comment is suspicious inside a linter code?
I feel like checking whether something is suspicious is a linter job, not the IR converter job.
We do mark strings as invalid as it would otherwise be impossible to report such errors, but this case doesn't look like exceptional enough.

src/phpdoc/parser.go
i582: Done.

src/ir/codegen/gen_equal.go
```diff
@@ -49,7 +49,7 @@ func (g *genEqual) writeCompare(w *bytes.Buffer, pkg *packageData, typ *typeData
 		case "[]ir.Node":
 			fmt.Fprintf(w, "    if !NodeSliceEqual(x.%[1]s, y.%[1]s) { return false }\n", field.Name())
 		case "ir.Doc":
-			fmt.Fprintf(w, "    if x.PhpDocComment != y.PhpDocComment { return false }\n")
+			fmt.Fprintf(w, "    if x.PhpDoc.Raw != y.PhpDoc.Raw { return false }\n")
```
quasilyte, line 52: (Not related to this PR, but still.)

By the way, it's a little strange that we generate comment-related comparison code for the `equal` method.
Nodes that are identical but have different comments should be equal.
I hope we're not using it when comparing an actual nodes.

src/ir/helper_types.go
```diff
@@ -20,6 +20,5 @@ type Class struct {
 }
 
 type Doc struct {
-	PhpDocComment string
-	PhpDoc        []phpdoc.CommentPart
+	phpdoc.Comment
 }
```
quasilyte, line 24: Maybe use `phpdoc.Comment` instead of `ir.Doc` directly and remove the `ir.Doc` type?

i582: Done.

quasilyte (approved):

PR #958, opened 2021-04-01
linter: remove code duplication in `Worker.IndexFile()` method

quasilyte:
Moved the repeated code to the new `parseWithCache()` method.


src/linter/worker.go
```diff
@@ -146,6 +146,14 @@ func (w *Worker) ParseContents(fileInfo workspace.FileInfo) (result ParseResult,
 	return result, nil
 }
 
+func (w *Worker) parseWithCache(cacheFileme string, file workspace.FileInfo) error {
```
quasilyte, line 149: Typo: cacheFileme -> cacheFilename

PR #962, opened 2021-04-04
ir/irconv: added lost colon token initialization for class method

i582:


quasilyte:
Is there a way to test it?
Maybe a printer test?

i582:
> Is there a way to test it?
> Maybe a printer test?

Printer tests do not use tokens, it seems then we will have to rewrite the printer completely.

quasilyte (approved):

PR #963, opened 2021-04-04
meta: added type inference for instance_cast

i582:
Since we now have two functions for `OverrideClassType` 
(`instance_deserialize` and `instance_cast`), the first of which 
must return a nullable type and the second not, a new type, 
`OverrideNullableClassType`, has been added that defines the type 
for functions that must return a nullable type described by the class name.

#977


src/meta/metainfo.go
```diff
@@ -85,6 +85,7 @@ const (
 // For example, \array_slice($arr) returns type of element (OverrideElementType) of the ArgNum=0
 type FuncInfoOverride struct {
 	OverrideType OverrideType
+	WithNull     bool // whenever we need to add null type (usually for OverrideClassType)
```
quasilyte, line 88: How about a new override type, OverrideNullableClassType.
You can share the code by moving the common logic to the function and add null type to the OverrideNullableClassType case.

i582: I thought about it, maybe this is really a better solution.

i582: Done.

src/linter/cache.go
```diff
@@ -44,8 +44,7 @@ import (
 //     46 - changed the way of inferring the return type of functions and methods
 //     47 - forced cache version invalidation due to the #921
 //     48 - renamed meta.TypesMap to types.Map; this affects gob encoding
-//     49 - added WithNull field to FuncInfoOverride structure
-const cacheVersion = 49
+const cacheVersion = 47
```
quasilyte: Why 47?

i582: It looks like this is a bug from the master. 
Let's increase the version to a separate PR.

quasilyte (approved):
Nice!

quasilyte:
By the way: do we want to have a few instance_deserialize tests as well?

i582:
> By the way: do we want to have a few instance_deserialize tests as well?

So we already have the `TestInstanceDeserializeType` tests.

PR #966, opened 2021-04-05
linter: added support for @noinspection PhpUnreachableStatementInspection 

i582:
`@noinspection PhpUnreachableStatementInspection` 
annotation disable deadCode checking.

### DIFF
--- a/src/linter/block.go
+++ b/src/linter/block.go
@@ -1,6 +1,7 @@
 package linter
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


src/linter/block.go
```diff
@@ -148,6 +149,11 @@ func (b *blockWalker) reportDeadCode(n ir.Node) {
 		return
 	}
 
+	if b.containsDisableInspection(n, "PhpUnreachableStatementInspection") {
+		b.ctx.deadCodeReported = true
+		return
+	}
```
quasilyte, line 155: I would move this `if` statement below the switch, closer to the actual `Report` call.

quasilyte: But it will probably make phpdoc matching harder as you'll have to save that comment somewhere else. Nevermind then.

quasilyte (approved):

Issue #332, opened 2020-01-10
Investigate Php Inspections from PhpStorm EA Extended


quasilyte:
https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-

PR #967, opened 2021-04-05
linter: added some checks from phpinspectionsea

i582:
Simplify
```php
$x[$i] !== null ? $x[$i] : $y;
// to
$x[$i] ?? $y;

array_key_exists($i, $x) ? $x[$i] : $y;
// to
$x[$i] ?? $y;

substr($x, $y, 1);
// to
$x[$y];

array_push($a, $v);
// to
$a[] = $v;
```

#332

### DIFF
--- a/src/cmd/embeddedrules/rules.php
+++ b/src/cmd/embeddedrules/rules.php
@@ -11,50 +11,50 @@
  * @after   $x ?: $y
  */
 function ternarySimplify() {
-    /**
-     * @maybe could replace the ternary with just $cond
-     * @type bool $cond
-     */
-    $cond ? true : false;
+  /**
+   * @maybe could replace the ternary with just $cond
+   * @type bool $cond
+   */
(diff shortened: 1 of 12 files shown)


src/cmd/embeddedrules/rules.php
```diff
@@ -11,30 +11,50 @@
...
-   * @pure $x
-   */
-  $x ? $x : $y;
-
-  /**
-   * @maybe could rewrite as `$x ?? $y`
-   * @fix $x ?? $y
-   */
-  isset($x) ? $x : $y;
+    /**
```
i582, line 14: It seems that auto-formatting happened here. Anyway, I think sooner or later we need to do formatting for the rules 😄

i582: Although this would cause merge conflicts, so reverted the formatting back.

quasilyte (approved):

PR #968, opened 2021-04-12
linter: improve oldStyleConstructor checker

i582:
Now the check does not work if there is a namespace. 
I think this can also be included in this PR, since this is 
also a kind of improvement 😄

### DIFF
--- a/src/linter/root.go
+++ b/src/linter/root.go
@@ -902,10 +902,6 @@ func (d *rootWalker) checkOldStyleConstructor(meth *ir.ClassMethodStmt) {
 	if !inClass {
 		return
 	}
-	_, containsConstruct := d.getClass().Methods.Get(`__construct`)
-	if containsConstruct {
-		return
-	}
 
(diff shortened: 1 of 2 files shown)


quasilyte:
If you want to add this change, please make this code trigger a method name case warning:
```
class Foo {
  public function foo() {} // it's old-style constructor
  public function__construct() {}
}
```
People may do this without knowing that `foo` will be considered to be a constructor, not a normal method.

I would actually not allow such "legacy compatibility" constructors at all. We don't support such old versions of PHP anyway in noverify. Our target is somewhere in PHP7-PHP8 range.

i582:
> I would actually not allow such "legacy compatibility" constructors at all. We don't support such old versions of PHP anyway in noverify. Our target is somewhere in PHP7-PHP8 range.

In that case, then it probably really doesn't make sense to have an 
exception for backward compatibility.

quasilyte (approved):

PR #969, opened 2021-04-12
cmd/embeddedrules: add strangeCast checker

i582:
Check for:

```php
$x . "";
$x . '';
"" . $x;
'' . $x;

0 + $x;
0.0 + $x;

+$x;
```


quasilyte:
There is one potential problem with `$x+0`.
Sometimes it's used for the code beauty:
```php
$a = $x + 0;
$b = $x + 1;
```
So the values for `$a` and `$b` are aligned.
`0 + $x` is less popular though, so it can be triggered as a strange cast.

quasilyte:
It can be safe to trigger a warning for `$x + 0` there is no `$x + 1` inside the current expression, previous and next statement.

i582:
> There is one potential problem with `$x+0`.
> Sometimes it's used for the code beauty:
> 
> ```
> $a = $x + 0;
> $b = $x + 1;
> ```
> 
> So the values for `$a` and `$b` are aligned.
> `0 + $x` is less popular though, so it can be triggered as a strange cast.

Then we can merge this PR without these patterns, and do separate processing for `$a + 0`.

quasilyte (approved):

quasilyte:
> Then we can merge this PR without these patterns, and do separate processing for $a + 0.

Sounds like a plan.

PR #973, opened 2021-04-14
linter: add finally return check for unreachable checker

i582:
Report the use of `return` in `finally` when
there are other `return`, `throw` in the `try-catch`.

The problem is they will be unreachable.

For example:

```php
function f(): int {
  try {
    throwException();
  } catch (Exception $_) {
    return 2; // unreachable
  } finally {
    return 1;
  }
}

f(); // always return 1
```

Report the use of `die` in `catch` when
there are `finally` block.

The problem is `finally` will be unreachable.

```php
function f(): int {
    try {
        throwException();
    } catch (RuntimeException $_) {
        return 2;
    } catch (Exception $_) {
        die();
    } finally { // block finally is unreachable (because 2 catch block contains a exit/die)
        return 1;
    }
}
```

#332


src/linter/block.go
```diff
@@ -723,16 +726,112 @@ func (b *blockWalker) handleTry(s *ir.TryStmt) bool {
...
 }
 
+func (b *blockWalker) checkUnreachableForFinallyReturn(tryStmt *ir.TryStmt, tryCtx *blockContext, finallyCtx *blockContext, catchContexts []*blockContext) {
+	if finallyCtx == nil {
+		return
+	}
+
+	var exitPoints []exitPoint
+
+	// if try contains some other return/die/etc statement
```
quasilyte, line 746: `s/if/If` + period in the end.

src/linter/block.go
```diff
@@ -723,16 +726,112 @@ func (b *blockWalker) handleTry(s *ir.TryStmt) bool {
...
+}
+
+type exitPoint struct {
+	n   ir.Node
+	typ string
+}
+
+func (b *blockWalker) findExitPointsByFlags(where ir.Node, exitFlags int) (points []exitPoint) {
+	irutil.Inspect(where, func(n ir.Node) bool {
+		if exitFlags&FlagReturn == FlagReturn {
```
quasilyte, line 795: Do you want to check that this bit is set?
Why not do it like this:
```go
exitFlags&FlagReturn != 0
```
It's less error-prone after the copy/paste.

src/linter/block.go
i582: Fixed.

src/tests/inline/testdata/checkers/finallyReturn.php
```diff
@@ -0,0 +1,130 @@
...
+        return 1;
+    }
+}
+
+function finallyReturnOkWithoutReturnInTryCatch(): int {
+    try {
+        throwException();
+    } catch (Exception $_) {
+    } finally {
+        return 1; /* ok, catch and try blocks don't contain return/exceptions/die/etc */
```
quasilyte, line 106: No reason to use `/**` comment here.

i582: Fixed.

src/linter/block.go
i582: Fixed.

quasilyte (approved):

PR #976, opened 2021-04-15
meta: add type inferring for not_null, not_false and create_vector from polyfills

i582:
Since type inference occurs before lazy types are resolved, 
this does not work entirely for function return types.

Perhaps we should add the ability to add a lazy type, which 
would say that when resolving types, need to remove a certain 
type, for example, `false` or `null`.

#977

### DIFF
--- a/src/types/typesmap.go
+++ b/src/types/typesmap.go
@@ -81,6 +81,18 @@ func (m Map) Map(fn func(string) string) Map {
 	return NewMapFromMap(mapped)
 }
 
+// Filter returns a new types map with the types of m for which fn returns true.
+// The result type map is never marked as precise.
+func (m Map) Filter(fn func(string) bool) Map {
+	mapped := make(map[string]struct{}, len(m.m))
+	for typ := range m.m {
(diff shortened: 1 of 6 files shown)


src/types/typesmap.go
```diff
@@ -81,6 +81,18 @@ func (m Map) Map(fn func(string) string) Map {
 	return NewMapFromMap(mapped)
 }
 
+// Filter returns a new types map with the types of m for which fn returns true.
+// The result type map is never marked as precise.
+func (m Map) Filter(fn func(string) bool) Map {
+	mapped := make(map[string]struct{}, len(m.m))
```
quasilyte, line 87: s/map/filtered/?

i582: Done,

quasilyte:
Conflicts.

i582:
> Conflicts.

Fixed.

i582:
Oh no...

quasilyte:
:sweat_smile:

PR #961, opened 2021-04-04
linter: added a check for 'array' type hint

i582:
For `array` type hint for 

1. parameters of functions and methods;
2. return type of functions and methods;
3. class properties.

now necessary to add the more precise type to `phpdoc`.

For example:

```php
function f(array $a) {} // bad

/**
 * @param int[] $a
 */
function f1(array $a) {} // ok
```

For mixed types, need to write `mixed[]`.

```php
/**
 * @param mixed[] $a
 */
function f1(array $a) {} // ok
```

> Closures are not checked and are skipped.

PR #979, opened 2021-04-16
linter: add check for parent typehint

i582:
Checks when type hint `parent` is used in a class that does not have a parent class.

Need to merged after #961.


quasilyte (approved):

src/tests/inline/testdata/checkers/typeHint.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+class Boo {}
+
+class FooBad {
+    public parent $a; // want `cannot use 'parent' typehint when current class has no parent`
+
+    /** */
```
quasilyte, line 8: We probably need some way to test one checker per file.
Like in run-filter-match tests.
Otherwise they're both brittle and require such phpdoc stubs.

PR #980, opened 2021-04-16
types: moved some things related to types from linter/types to types

i582:


src/types/lazy.go
```diff
@@ -329,10 +247,84 @@ func FormatType(s string) (res string) {
 	return "unknown(" + s + ")"
 }
 
-func IsClassType(s string) bool {
-	return strings.HasPrefix(s, `\`) && !IsShapeType(s) && !IsArrayType(s)
+func slice(typ byte, byteFields []uint8, args ...string) []byte {
```
i582, line 250: Moved to the end of the file so that there are exported functions at the beginning of the file.

src/linter/block_linter.go
```diff
@@ -1145,10 +1145,6 @@ func (b *blockLinter) checkFormatString(e *ir.FunctionCallExpr, arg *ir.Argument
 	}
 }
 
-func (b *blockLinter) isArrayType(typ types.Map) bool {
```
i582, line 1148: If we refactor the types in the future, I don't really want to 
have such functions somewhere outside the `types` package, 
since they then need to be looked for, so I moved them to 
the `types.Map` method.

src/linter/root.go
```diff
@@ -603,7 +603,7 @@ func (d *rootWalker) handleFuncStmts(params []meta.FuncParam, uses, stmts []ir.N
 	switch {
 	case b.bareReturn && b.returnsValue:
 		b.returnTypes = types.MergeMaps(b.returnTypes, types.NullType)
-	case b.returnTypes.IsEmpty() && b.returnsValue:
+	case b.returnTypes.Empty() && b.returnsValue:
```
i582, line 606: Just for the consistency.

src/linter/types.go
```diff
@@ -263,18 +263,12 @@ func (n typeNormalizer) NormalizeTypes(typeList []types.Type) {
 	}
 }
 
-func (n typeNormalizer) string2name(s string) *ir.Name {
```
i582, line 266: The function has been used once and such conversions 
are usually done in place, so there is no point in this function.

src/solver/exprtype.go
```diff
@@ -197,7 +197,7 @@ func binaryPlusOpType(sc *meta.Scope, cs *meta.ClassParseState, left, right ir.N
 	// TODO: PHP will raise fatal error if one operand is array and other is not, so we may check it too
 	leftType := ExprTypeLocalCustom(sc, cs, left, custom)
 	rightType := ExprTypeLocalCustom(sc, cs, right, custom)
-	if leftType.IsArray() && rightType.IsArray() {
+	if leftType.IsLazyArray() && rightType.IsLazyArray() {
```
i582, line 200: All methods that somehow work with lazy types must have the word `Lazy`.

src/types/map.go
```diff
@@ -37,6 +37,125 @@ type Map struct {
 	m     map[string]struct{}
 }
 
+// MergeMaps creates a new types map from union of specified type maps
```
i582, line 40: Everything related to creation has been moved up and grouped.

quasilyte:
Is it possible to send refactoring first and move things in a separate PR?
It would be easier to review in that way.

i582:
> Is it possible to send refactoring first and move things in a separate PR?
> It would be easier to review in that way.

Yep.

src/types/map.go
i582: Revert.

src/types/lazy.go
i582: Revert.

quasilyte (approved):

PR #982, opened 2021-04-19
linter: fix panic if no token for node

i582:


i582:
@quasilyte see it please, there is a small, but important, change.

quasilyte (approved):

PR #991, opened 2021-05-23
linter: added varShadow checker

i582:
`varShadow` checks that the variables from the
key and the variable in foreach did not shadow
the parameters of the current function.

### DIFF
--- a/src/linter/block_linter.go
+++ b/src/linter/block_linter.go
@@ -156,7 +156,7 @@ func (b *blockLinter) enterNode(n ir.Node) {
 		b.walker.r.checkKeywordCase(n, "else")
 
 	case *ir.ForeachStmt:
-		b.walker.r.checkKeywordCase(n, "foreach")
+		b.checkForeach(n)
 	case *ir.ForStmt:
 		b.walker.r.checkKeywordCase(n, "for")
 	case *ir.WhileStmt:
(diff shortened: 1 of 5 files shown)


src/linter/block_linter.go
```diff
@@ -184,6 +184,39 @@ func (b *blockLinter) enterNode(n ir.Node) {
 	}
 }
 
+func (b *blockLinter) checkForeach(n *ir.ForeachStmt) {
+	b.walker.r.checkKeywordCase(n, "foreach")
+
+	var vars []*ir.SimpleVar
+
+	findAllVars := func(n ir.Node) bool {
```
quasilyte, line 192: s/findAllVars/collectAllVars/?

src/linter/block_linter.go
```diff
@@ -184,6 +184,39 @@ func (b *blockLinter) enterNode(n ir.Node) {
...
+		irutil.Inspect(n.Key, findAllVars)
+	}
+
+	fun, ok := b.walker.r.currentFunction()
+	if !ok {
+		return
+	}
+
+	for _, v := range vars {
+		for _, param := range fun.Params {
```
quasilyte, line 212: I'm OK with a nested loop here as it will never be a bottleneck, probably.
But we have functions with 1k+ lines of code, many foreach loops (and probably many params).

How about creating a map inside a walker when we enter a function that would hold all params?
It should probably be lazily initialized, the first time we call `getCurrentFuncParams()` method.

quasilyte (approved):

PR #996, opened 2021-06-17
linter: added check for repeated nullable type

i582:
Also fixed a bug where a nullable field in shape
caused the entire shape to become nullable.

### DIFF
--- a/src/linter/types.go
+++ b/src/linter/types.go
@@ -23,7 +23,19 @@ func typesFromPHPDoc(ctx *rootContext, typ phpdoc.Type) ([]types.Type, warningSt
 	conv := phpdocTypeConverter{ctx: ctx}
 	result := conv.mapType(typ.Expr)
 	if conv.nullable {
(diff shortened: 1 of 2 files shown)


src/linter/types.go
```diff
@@ -23,7 +23,19 @@ func typesFromPHPDoc(ctx *rootContext, typ phpdoc.Type) ([]types.Type, warningSt
 	conv := phpdocTypeConverter{ctx: ctx}
 	result := conv.mapType(typ.Expr)
 	if conv.nullable {
-		result = append(result, types.Type{Elem: "null"})
+		alreadyHasNull := false
+
+		for _, tp := range result {
+			if tp.Elem == "null" {
+				alreadyHasNull = true
+				conv.warning = "twice nullable doesn't make sense"
```
quasilyte, line 31: s/twice/repeated/?

i582: Fixed.

quasilyte (approved):

PR #1002, opened 2021-06-29
linter: added `*FromAnnotation` value for `meta.PropertyFlags` and `meta.FuncFlags`

i582:
These values are needed in order to distinguish between
methods and properties that are actually in the class and
which are defined in the annotation for the class.

### DIFF
--- a/src/meta/metainfo.go
+++ b/src/meta/metainfo.go
@@ -75,9 +75,10 @@ type FuncInfo struct {
 	Doc          PhpDocInfo
 }
 
-func (info *FuncInfo) IsStatic() bool   { return info.Flags&FuncStatic != 0 }
-func (info *FuncInfo) IsAbstract() bool { return info.Flags&FuncAbstract != 0 }
-func (info *FuncInfo) IsPure() bool     { return info.Flags&FuncPure != 0 }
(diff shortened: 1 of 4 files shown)


src/meta/metainfo.go
```diff
@@ -66,9 +75,10 @@ type FuncInfo struct {
 	Doc          PhpDocInfo
 }
 
-func (info *FuncInfo) IsStatic() bool   { return info.Flags&FuncStatic != 0 }
-func (info *FuncInfo) IsAbstract() bool { return info.Flags&FuncAbstract != 0 }
-func (info *FuncInfo) IsPure() bool     { return info.Flags&FuncPure != 0 }
+func (info *FuncInfo) IsStatic() bool       { return info.Flags&FuncStatic != 0 }
+func (info *FuncInfo) IsAbstract() bool     { return info.Flags&FuncAbstract != 0 }
+func (info *FuncInfo) IsPure() bool         { return info.Flags&FuncPure != 0 }
+func (info *FuncInfo) FromAnnotation() bool { return info.Flags&FuncFromAnnotation != 0 }
```
quasilyte, line 81: IsFromAnnotation?

i582: Fixed.

quasilyte (approved):

PR #1005, opened 2021-07-01
linter: added check that null is assigned to not nullable property

i582:


quasilyte:
I would suggest adding some variety in tests.
At least 1 static property would be nice, for instance.

src/linter/root.go
```diff
@@ -829,30 +829,83 @@ func (d *rootWalker) enterPropertyList(pl *ir.PropertyListStmt) bool {
...
 	}
 
 	return true
 }
 
+func (d *rootWalker) checkAssignNullToNotNullableProperty(prop *ir.PropertyStmt, propTypes types.Map) {
+	assignNull := false
+
+	if expr, ok := prop.Expr.(*ir.ConstFetchExpr); ok {
+		assignNull = expr.Constant.Value == "null"
```
quasilyte, line 866: Just curious, will it work with `NULL`?
We can use `strings.EqualFold` here, I think.

src/linter/report.go
```diff
@@ -703,6 +703,25 @@ function f(array $a) {}`,
   foreach ([1, 2] as $b) {
     echo $b;
   }
+}`,
+		},
+
+		{
+			Name:     "nullProp",
```
quasilyte, line 710: `propNullDefalt`?

src/linter/root.go
```diff
@@ -829,30 +829,83 @@ func (d *rootWalker) enterPropertyList(pl *ir.PropertyListStmt) bool {
...
+	}
+
+	if assignNull && !propTypes.Empty() {
+		onlyClasses := true
```
quasilyte, line 870: I would add a test with something like `int|Foo` where we mix classes and non-classes.

src/linter/root.go
```diff
@@ -829,30 +829,83 @@ func (d *rootWalker) enterPropertyList(pl *ir.PropertyListStmt) bool {
...
+	}
+
+	from := prop.Position.StartPos
+	to := prop.Position.EndPos
+	have := d.file.Contents()[from:to]
+	start := bytes.IndexByte(have, '=')
+	if start == -1 {
+		return
+	}
+	if have[start-1] == ' ' {
```
quasilyte, line 897: What if it has several spaces before `=`?
```
public $foo  = 1;
```
We can add this test just in case.
I don't think that it's necessary to fix the formatting there, but it's still better to cover such a case.
Especially since we covered no spaces form `x=y`.

src/linter/root.go
```diff
@@ -829,30 +829,83 @@ func (d *rootWalker) enterPropertyList(pl *ir.PropertyListStmt) bool {
...
+
+func (d *rootWalker) addFixForNullForNotNullableProperty(prop *ir.PropertyStmt) {
+	if !d.config.ApplyQuickFixes {
+		return
+	}
+
+	from := prop.Position.StartPos
+	to := prop.Position.EndPos
+	have := d.file.Contents()[from:to]
+	start := bytes.IndexByte(have, '=')
```
quasilyte, line 893: I think that it's not a start, but rather an `end` as we trim until that position.
It's not a start of `=` too as we also move backward to remove extra spaces.

src/linter/root.go
i582: Fixed.

src/linter/root.go
i582: Added.

src/linter/root.go
i582: Fixed.

src/linter/root.go
i582: Fixed.

src/linter/report.go
i582: Renamed.

quasilyte (approved):

PR #1027, opened 2021-07-04
cmd,linter: filtered reports inside the `Report` function

i582:
Now, if the check is not enabled, the report will not be added in `Report` function.

### DIFF
--- a/src/cmd/main.go
+++ b/src/cmd/main.go
@@ -337,9 +337,6 @@ func analyzeReports(l *linterRunner, cfg *MainConfig, diff []*linter.Report) (cr
 		if cfg.BeforeReport != nil && !cfg.BeforeReport(r) {
 			continue
 		}
-		if !l.checks.IsEnabledReport(r) {
-			continue
-		}
 
 		filtered = append(filtered, r)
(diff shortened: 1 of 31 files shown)


src/cmd/main.go
```diff
@@ -330,21 +330,20 @@ func haveAutofixableReports(config *linter.Config, reports []*linter.Report) boo
...
 
-func analyzeReports(l *linterRunner, cfg *MainConfig, diff []*linter.Report) (criticalReports int, containsAutofixableReports bool) {
+func analyzeReports(l *linterRunner, cfg *MainConfig, diff []*linter.Report) (criticalReports, minorReports int, containsAutofixableReports bool) {
 	filtered := make([]*linter.Report, 0, len(diff))
 
 	for _, r := range diff {
 		if cfg.BeforeReport != nil && !cfg.BeforeReport(r) {
 			continue
 		}
-		if !l.IsEnabledReport(r) {
```
i582, line 340: We no longer need to be checked here, as it is checked when creating a report.

quasilyte (approved):

Issue #1037, opened 2021-07-06
Smartcasts


i582:
This issue tracks the status of smart casts in the linter.

## `instanceof`
Only works for the simple case:
```php
if ($a instanceof Foo) {
   // $a has \Foo type
}
```
And at the moment the variable continues to have this type after the if block, which is a bug.

PR #1038, opened 2021-07-07
linter: initial support of complex instanceof for if/else and ternary

i582:
## Description
TODO

#1037 
#363


quasilyte (changes requested):
Please move closure improvements to another PR.
We can merge them right away then.

quasilyte:
We discussed some `!` nesting.
If you think that it's not an issue, feel free to merge.

PR #1039, opened 2021-07-07
phpdoctypes: for `iterable<T>` now will be assume type `array<T>`

i582:


quasilyte:
I am not sure you can use iterable in any array context.
Can you apply indexing to iterable?

What is the motivation behind this change?

PR #1041, opened 2021-07-07
linter: now, if the function returns array, then inferred type is taken into account

i582:
If the resulting type is `mixed[]`, then if the actual type is a specific
array type, then we use it, otherwise, we combine this type with the
resulting `mixed[`] type.

If a new element is added to an empty array, then the `empty_array` type
is removed.

#849


quasilyte:
No tests?

PR #1043, opened 2021-07-07
linter: improved error output in PHPDoc

i582:
The line on which the error is now is the context line.
Also, the underline now highlights the desired part, rather than
the entire line (for functions, for example).

Before:
```
WARNING phpdocLint: @param for non-existing argument $package at testdata/options-resolver/OptionsResolver.php:428
    public function setDeprecated(string $option/*, string $package, string $version*/): self
                    ^^^^^^^^^^^^^
```

After:

```
WARNING phpdocLint: @param for non-existing argument $package at testdata/options-resolver/OptionsResolver.php:424
     * @param string          $package The name of the composer package that is triggering the deprecation
                              ^^^^^^^^
```

### DIFF
--- a/src/linter/phpdoc_util.go
+++ b/src/linter/phpdoc_util.go
@@ -13,10 +13,26 @@ import (
 )
 
 type PHPDocPlace struct {
-	Node ir.Node
-	Line int
-	Part int
-	All  bool
+	Node  ir.Node
+	Line  int
+	Field int
+	All   bool
+}
+
+func PHPDocLine(n ir.Node, line int) PHPDocPlace {
+	return PHPDocPlace{
+		Node: n,
+		Line: line,
+		All:  true,
+	}
+}
+
+func PHPDocLineField(n ir.Node, line int, field int) PHPDocPlace {
+	return PHPDocPlace{
+		Node:  n,
+		Line:  line,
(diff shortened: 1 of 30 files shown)


src/linter/phpdoc_util.go
```diff
@@ -12,27 +12,61 @@ import (
 	"github.com/VKCOM/noverify/src/types"
 )
 
-type phpdocErrors struct {
-	phpdocLint []string
-	phpdocType []string
+type PHPDocPlace struct {
```
quasilyte, line 15: Maybe we can move it to `phpdoc` package?
Along with related functions like PHPDocLine.

Also, Place should probably be called Location.

src/linter/phpdoc_util.go
```diff
@@ -12,27 +12,61 @@ import (
...
+	Node  ir.Node
+	Line  int
+	Field int
+	All   bool
```
quasilyte, line 19: What does "All" mean here?

i582: Renamed with `WholeLine`.

src/linter/phpdoc_util.go
i582: Done.

PR #1064, opened 2021-07-16
linter: added `argsReverse` checker

i582:
It checks if parameters are passed in reverse order.
For functions with two parameters only.


quasilyte:
`argsReverse` sounds like an actions.
Diagnostic names should try to be noun-like.
`reversedArgs` is probably not the best option thought.
I would use something like `argsOrder`, but maybe there is even a better name for it.

PR #1167, opened 2021-11-01
linter: added initial support for `@noverify-suppress` annotations

i582:
The syntax for the annotation is as follows:

```php
/** @noverify-suppress <checker-name> or all */
```

If `all` is specified, then all warnings in the expression or 
statement inside the function will be suppressed.

This annotation can only be used for expressions in the global 
scope and for expressions and statements in the function scope.


YuriyNasretdinov:
If you are going to do this (which I don't agree that is a good idea, but that's a separate discussion), I would suggest to make syntax compatible with PHPStorm and other linters as much as possible.

i582:
> If you are going to do this (which I don't agree that is a good idea, but that's a separate discussion), I would suggest to make syntax compatible with PHPStorm and other linters as much as possible.

As for compatibility, everything is rather complicated, because both PHPStorm and Psalm use their own check names and their annotations, it seems impossible to make it compatible with them.

As for the need... I kind of agree that this is not a good idea, but as a default linter feature it's not bad. In the our code it will be allowed only for a few checks, which often break push when refactoring (for example, the `deprecated` check).

PR #1175, opened 2022-04-17
Add stdout report output

Danil42Russia:
Added report output selection


i582:
Please explain why this is necessary?

i582 (approved):

PR #1192, opened 2022-08-18
added PhpStorm plugins

Danil42Russia:


docs/phpstorm-plugin.md
```diff
@@ -0,0 +1,12 @@
+# Using NoVerify for PhpStorm
```
i582, line 1: maybe `in` instead of `for`?

PR #1213, opened 2024-02-19
linter: new rule for declare(strict_types)

Hidanio:
- added new linter rule: declareStrictTypes can not be "0"
- added quickFix for declareStrictTypes
- added new linter rule: lack of declare(strict_types=1) section
-  added quickFix for lack of declare(strict_types=1) section
- added & updated tests
- updated doc
- new rules for .gitignore
- refactored warnings


docs/install.md
```diff
@@ -73,13 +73,13 @@ noverify version
 
 You're done! Proceed to the [Getting started](/docs/getting_started.md) page.
 
-## With `go get`
+## With `go install`
```
i582, line 76: these changes should be in a separate PR as they do not relate to its main topic

src/tests/golden/testdata/inflector/lib/Doctrine/Inflector/CachedWordInflector.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-declare(strict_types=1);
+declare(strict_types = 1);
```
i582, line 3: changes to test data formatting only pollute the PR, making it difficult to review

src/linter/worker.go
```diff
@@ -315,6 +315,11 @@ func (w *Worker) analyzeFile(file *workspace.File, rootNode *ir.Root) (*rootWalk
 	}
 	walker.afterLeaveFile()
 
+	if !walker.declareSection {
+		walker.Report(rootNode, LevelWarning, "noDeclareSection", "You have no declare section with strict_types=1 argument")
```
i582, line 319: maybe "missed `declare(strict_types=1)` directive"?

docs/install.md
Hidanio: I am totally agree with you (and that was the original title of this PR), but we had an agreement with @Danil42Russia  that this PR would also include changes to the documentation, corrections to some warnings, and test data. But if you insist, I can separate them

src/tests/golden/testdata/inflector/lib/Doctrine/Inflector/CachedWordInflector.php
Hidanio: Reply is the same as to the previous comment

src/linter/worker.go
Hidanio: I used a more formalized form, I agree, it could be simplified

docs/install.md
i582: ok!

PR #1215, opened 2024-03-13
linter: fixed unboxing from null for arrays

Hidanio:


YuriyNasretdinov:
Seems quite non-obvious to me why does UnwrapArray actually panic in this case...

Hidanio:
> Seems quite non-obvious to me why does UnwrapArray actually panic in this case...

@YuriyNasretdinov  The problem in unwrap1 function:
```
const stringLenBytes = 4

func unwrap1(s string) (one string) {
	return s[stringLenBytes+1:] // do not care about length, there is only 1 param
}
```
if we have "null" string, then we will have slice from 5, but we have only 3 indexes. we can modify unwrap1 function to return empty string if len(s) > stringLenBytes+1, but we are talking about "types" and "null" is a special type that`s why i think we should add condition in logic of LazyArrayElemType. What do you think?

YuriyNasretdinov:
Thanks for the explanation. I think it makes sense now. BTW I am not a maintainer of the project (but I was at some point), so I'm just being curious :).

src/types/map.go
```diff
@@ -410,6 +410,10 @@ func (m Map) LazyArrayElemType() Map {
 
 	mm := make(map[string]struct{}, m.Len())
 	for typ := range m.m {
+		if typ == "null" {
```
Danil42Russia, line 413: As a crutch OK :(

But we should add a TODO comment that there should be an inspection here that only iterators can be unpacked.

src/types/map.go
```diff
@@ -420,7 +424,6 @@ func (m Map) LazyArrayElemType() Map {
 			// is a more precise type.
 			continue
 		}
-
```
Danil42Russia, line 423: why?

src/tests/regression/issue1214_test.go
```diff
@@ -0,0 +1,17 @@
+package regression_test
+
+import (
+	"testing"
+
+	"github.com/VKCOM/noverify/src/linttest"
+)
+
+func TestIssue1214(t *testing.T) {
```
Danil42Russia, line 9: The test must be in `TestArrayTypesWithUnpack`

src/types/map.go
Danil42Russia, line 1: 👀

Hidanio: poshel nahui

src/types/map.go
```diff
@@ -410,6 +410,11 @@ func (m Map) LazyArrayElemType() Map {
 
 	mm := make(map[string]struct{}, m.Len())
 	for typ := range m.m {
+		// TODO: https://github.com/VKCOM/noverify/issues/1227
+		if typ[0] >= WMax {
```
Danil42Russia, line 414:
```go
typ == "" || typ[0] >= WMax
```

And it's better to make it a function. And replace this pattern with this function :)

src/solver/solver.go
```diff
@@ -20,14 +20,16 @@ func mixedType() map[string]struct{} {
 }
 
 // resolveType resolves function calls, method calls and global variables.
-//   curStaticClass is current class name (if inside the class, otherwise "")
+//
+//	curStaticClass is current class name (if inside the class, otherwise "")
```
Danil42Russia, line 24: It's definitely unnecessary

src/tests/exprtype/exprtype_test.go
```diff
@@ -2201,6 +2201,44 @@ function f() {
...
+$b2 = [...$a2];
+exprtype($b2, "mixed");
+
+// with bool type
+$a3 = true;
+$b3 = [...$a3];
+exprtype($b3, "mixed");
+
+// null
+ $a4 = null;
```
Danil42Russia, line 2228: formatting has gone (fix it everywhere with php

Danil42Russia (approved):

PR #1217, opened 2024-03-22
linter: new rule for dangerous bool/int condition

Hidanio:
Sometimes during debugging we can write this kind of code
```
if(1){
}
while(true){
}
```
This code can be very dangerous.

Added:

- new rule for: if/while/do statemenst
- tests for it


Hidanio:
I think we should change Warning -> notice

src/linter/block_linter.go
Danil42Russia, line 1: what about `ElseIfStmt`?

Hidanio: Good catch. I will add it

Hidanio: Done

Danil42Russia (approved):

PR #1218, opened 2024-04-01
linter: new rule about deprecated string interpolation

Hidanio:


src/linter/block_linter.go
```diff
@@ -198,6 +202,18 @@ func (b *blockLinter) enterNode(n ir.Node) {
 	}
 }
 
+func (b *blockLinter) checkStringInterpolationDeprecation(str *ir.Encapsed) {
+	for _, item := range str.Parts {
+		variable, ok := item.(*ir.SimpleVar)
+		if ok {
+			if variable.IdentifierTkn.Value[0] != 36 { // 36 is $
```
YuriyNasretdinov, line 209: Can't you use `'$'` instead?

PR #1219, opened 2024-05-03
linter: new inspection for php functions aliases

Hidanio:


src/php/php_aliases.go
```diff
@@ -0,0 +1,123 @@
+package php
+
+var Aliases = map[string]string{
```
Danil42Russia, line 3: In `src/ir/phpcore/phpcore.go` it has this. Let's summarize this...

Hidanio: Fixed

PR #1222, opened 2024-06-18
dynamic rules: `path-exclude` rule fix

Hidanio:


Danil42Russia (commented):
Why prefix `dinamic rules` (with an error) and not just `rules`?

Hidanio:
> Why prefix `dinamic rules` (with an error) and not just `rules`?

Because this two different conceptions and we should not mix it. We have different logic for launch dynamic rules and embedded (I know that dynamic rules are called that way in the code and I think it’s wrong). We must forcefully separate this concept

src/tests/rules/rules_test.go
```diff
@@ -488,3 +488,51 @@ function type_type_check(string $animal_name, int $animal_id) {
...
+          $hello = 'echo 123;';
+          eval($hello);
+          eval('echo 456;');
+        `
+	test.AddNamedFile("www/no", code)
+
+	test.RunRulesTest()
+}
+
+func TestRulePathExclude2(t *testing.T) {
```
Danil42Russia, line 514: Don't give incremental tests. Give a normal name.

Hidanio: Renamed

Danil42Russia (approved):

Issue #1208, opened 2023-01-23
Multiple `@path` directives


vkaverin:
**Is your feature request related to a problem? Please describe.**

Currently, it's forbidden to put multiple `@path`s for single rule. Rule like

```php
/**
  * @warning Do not use `foo()`!
  * 
  * @path www/A/
  * @path www/B/
  */
foo();
```

will end up with `duplicate @path constraint` error.

So, to apply rule `foo()` for both `www/A/` and `www/B/` it's necessary to copy-paste the rule for them individually:

```php
/**
  * @warning Do not use `foo()`!
  * 
  * @path www/A/
  */
foo();

/**
  * @warning Do not use `foo()`!
  * 
  * @path www/B/
  */
foo();
```

**Describe the solution you'd like**

Allow to specify `@path`s and merge them.

PR #1224, opened 2024-07-11
dinamic rules: multiple path

Hidanio:
More about feature [here](https://github.com/VKCOM/noverify/issues/1208).

### DIFF
--- a/src/linter/utils.go
+++ b/src/linter/utils.go
@@ -68,15 +68,15 @@ func mergeTypeMaps(left types.Map, right types.Map) types.Map {
 // The types are inferred as follows:
 // 1. If there is a @return annotation, then its value becomes the return type;
 //
-// 2. If there is a type hint, then it is added to the types from the @return.
-//    If the @return is empty, then the type matches the type hint itself;
+//  2. If there is a type hint, then it is added to the types from the @return.
+//     If the @return is empty, then the type matches the type hint itself;
 //
-// 3. If the resulting type is mixed[], then if the actual type is a specific
-//    array type, then we use it, otherwise we combine this type with the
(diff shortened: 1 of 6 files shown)


src/linter/utils.go
```diff
@@ -494,10 +494,28 @@ func cloneRulesForFile(filename string, ruleSet *rules.ScopedSet) *rules.ScopedS
...
-			if !strings.Contains(filename, rule.Path) || isFilePathExcluded(filename, rule) {
+			if isFilePathExcluded(filename, rule) {
 				continue
 			}
-			res = append(res, rule)
+
+			match := false
+
+			if rule.Paths == nil {
+				if strings.Contains(filename, "") {
```
Danil42Russia, line 504: doesn't make sense, will always be true

Hidanio: Thank you. Removed

src/linter/utils.go
```diff
@@ -494,10 +494,20 @@ func cloneRulesForFile(filename string, ruleSet *rules.ScopedSet) *rules.ScopedS
...
 		res := make([]rules.Rule, 0, len(ruleByKind))
 		for _, rule := range ruleByKind {
-			if !strings.Contains(filename, rule.Path) || isFilePathExcluded(filename, rule) {
+			if isFilePathExcluded(filename, rule) {
 				continue
 			}
-			res = append(res, rule)
+
+			if rule.Paths == nil {
+				res = append(res, rule)
```
Danil42Russia, line 502: append `continue`

Hidanio: Fixed

Danil42Russia (approved):

PR #1228, opened 2024-07-25
linter: new rule for declare(strict_types)

Hidanio:
- added new linter rule: declareStrictTypes can not be "0"
- added quickFix for declareStrictTypes
- added new linter rule: lack of declare(strict_types=1) section
- added quickFix for lack of declare(strict_types=1) section

### DIFF
--- a/src/linter/worker.go
+++ b/src/linter/worker.go
@@ -316,7 +316,7 @@ func (w *Worker) analyzeFile(file *workspace.File, rootNode *ir.Root) (*rootWalk
 	walker.afterLeaveFile()
 
 	if !walker.declareSection {
-		walker.Report(rootNode, LevelWarning, "noDeclareSection", "You have no declare section with strict_types=1 argument")
+		walker.Report(rootNode, LevelWarning, "noDeclareSection", "Missed declare(strict_types=1) directive")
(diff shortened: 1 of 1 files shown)


src/linter/root.go
```diff
@@ -69,10 +69,12 @@ type rootWalker struct {
 	// name matches the pattern and @linter disable was encountered
 
 	// strictTypes is true if file contains `declare(strict_types=1)`.
-	strictTypes bool
-	strictMixed bool
+	strictTypes    bool
+	strictMixed    bool
+	declareSection bool
```
Danil42Russia, line 74: I propose to do it differently. We have an `ir.Root` node that describes the entire file. Essentially, we can check if there is an `ir.DeclareStmt` at its level, and if there is, show an error.

see `block.go`

src/linter/report.go
```diff
@@ -25,6 +25,24 @@ func addBuiltinCheckers(reg *CheckersRegistry) {
 			After:    `$s = strip_tags($s, '<br>')`,
 		},
 
+		{
+			Name:     "notStrictTypes",
```
Danil42Russia, line 29: Let's split the checks into two different MRs, shall we?

PR #1231, opened 2024-07-27
updated `golangci-lint` and fixed errors in the code

Danil42Russia:


src/phpgrep/tracing_disabled.go
```diff
@@ -1,3 +1,4 @@
+//go:build !tracing
```
Hidanio, line 1: space

src/phpgrep/tracing_enabled.go
```diff
@@ -1,3 +1,4 @@
+//go:build tracing
```
Hidanio, line 1: space

Hidanio (changes requested):

src/phpgrep/tracing_disabled.go
Danil42Russia: no

src/phpgrep/tracing_enabled.go
Danil42Russia: no

Hidanio (approved):

PR #1232, opened 2024-07-28
dynamic rules: new `@link` tag

Hidanio:


src/linter/cache.go
```diff
@@ -16,40 +16,42 @@ import (
...
+//	    46 - changed the way of inferring the return type of functions and methods
+//	    47 - forced cache version invalidation due to the #921
+//	    48 - renamed meta.TypesMap to types.Map; this affects gob encoding
+//	    49 - for shape, names are now generated using the keys that make up this shape
+//	    50 - added Flags field for meta.PropertyInfo
+//	    51 - added anonymous classes
+//	    52 - renamed all PhpDoc and Phpdoc with PHPDoc
+//	    53 - added DeprecationInfo for functions and methods and support for some attributes
+//	    54 - forced cache version invalidation due to the #1165
+//		55 - added @link tag -> new field in Rule
```
Danil42Russia, line 54: it's not like you changed the meta information, why upgrade the version?

src/rules/parser.go
```diff
@@ -177,6 +177,16 @@ func (p *parser) parseRuleInfo(st ir.Node, labelStmt ir.Node, proto *Rule) (Rule
 			}
 			rule.Name = part.Params[0]
 
+		case "link":
+			if len(part.Params) != 1 {
+				return rule, p.errorf(st, "@link expects exactly 1 param, got %d", len(part.Params))
+			}
+			var link = part.Params[0]
+			if !isURL(link) {
```
Danil42Russia, line 185: you've really overcomplicated the logic. Why is there a check for `url`?

src/rules/parser.go
```diff
@@ -339,6 +348,11 @@ func (p *parser) parseRuleInfo(st ir.Node, labelStmt ir.Node, proto *Rule) (Rule
 	return rule, nil
 }
 
+func isURL(str string) bool {
+	re := regexp.MustCompile(`^((https?|ftp)://)?[^\s/$.?#].\S*$`)
```
Danil42Russia, line 352: the regular expression is definitely not correct.

src/rules/parser.go
Hidanio: After discussion, I removed it completely

src/rules/parser.go
Hidanio: After discussion, I removed it completely

PR #1237, opened 2024-10-22
linter: base system for on path rule filtering

Hidanio:
tied with [prototype](https://github.com/VKCOM/noverify/pull/1236).

### DIFF
--- a/src/linter/conf.go
+++ b/src/linter/conf.go
@@ -77,21 +77,21 @@ type Config struct {
 
 type RuleNode struct {
 	Children map[string]*RuleNode // Child nodes (subdirectories and files)
-	Enable   map[string]bool      // Rules enabled at this level
-	Disable  map[string]bool      // Disabled rules at this level
+	Enabled  map[string]bool      // Rules enabled at this level
+	Disabled map[string]bool      // Disabled rules at this level
 }
 
 func NewRuleNode() *RuleNode {
 	return &RuleNode{
 		Children: make(map[string]*RuleNode),
-		Enable:   make(map[string]bool),
-		Disable:  make(map[string]bool),
+		Enabled:  make(map[string]bool),
+		Disabled: make(map[string]bool),
 	}
 }
 
 type PathRuleSet struct {
-	Enable  map[string]bool // Rules that are enabled for this path
-	Disable map[string]bool // Rules that are disabled for this path
+	Enabled  map[string]bool // Rules that are enabled for this path
+	Disabled map[string]bool // Rules that are disabled for this path
 }
 
 func BuildRuleTree(pathRules map[string]*PathRuleSet) *RuleNode {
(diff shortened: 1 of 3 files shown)


src/linter/conf.go
```diff
@@ -68,6 +75,82 @@ type Config struct {
...
+		for rule := range ruleSet.Disable {
+			currentNode.Disable[rule] = true
+		}
+	}
+
+	return root
+}
+
+func IsRuleEnabled(root *RuleNode, filePath string, checkRule string) bool {
+	normalizedPath := filepath.ToSlash(filepath.Clean(filePath))
```
Danil42Russia, line 127: 1. let's move it to `root.go`
2. Can we definitely have a situation where we need to apply `filepath.Clean`?

src/linter/root.go
```diff
@@ -1556,6 +1556,25 @@ func (d *rootWalker) Report(n ir.Node, level int, checkName, msg string, args ..
 		}
 	}
 
+	var rootNode = d.config.PathRules
```
Danil42Russia, line 1559: after `if`

src/linter/conf.go
```diff
@@ -68,6 +75,82 @@ type Config struct {
...
+		}
+		for rule := range ruleSet.Disable {
+			currentNode.Disable[rule] = true
+		}
+	}
+
+	return root
+}
+
+func IsRuleEnabled(root *RuleNode, filePath string, checkRule string) bool {
```
Danil42Russia, line 126: `IsRuleEnabledForPath`

src/linter/conf.go
```diff
@@ -68,6 +75,82 @@ type Config struct {
...
+	normalizedPath := filepath.ToSlash(filepath.Clean(filePath))
+	parts := strings.Split(normalizedPath, "/")
+	currentNode := root
+
+	// Starting with global state. We have guarantee while parsing config that rule is `on` and exist
+	ruleState := true
+
+	for _, part := range parts {
+		if part == "" {
```
Danil42Russia, line 135: is this exactly possible?

src/linter/root.go
Hidanio: Good catch. Done

src/linter/conf.go
Hidanio: Renamed

src/linter/conf.go
Hidanio: 1. Done
2. I think that this is a good practice to combine it:
2.1. filePath := "A//B///C/file.php" will give us parts = ["A", "", "B", "", "", "C", "file.php"] => this is not very good
2.2. filePath := "A/B/C/" => parts = ["A", "B", "C", ""]
Or some another examples. I think that is ok to use it

src/linter/conf.go
Hidanio: I think yes:
1. filePath := "/A/B/C" => ["", "A", "B", "C"]
2. filePath := "A//B/C" and "A/B/C/". We used filepath.Clean, but if something will be wrong...idkn
I think that enabling this check ensures that our code will correctly handle any paths, regardless of their origin or format

src/linter/conf.go
Danil42Russia: 1. I don't see it( `filePath = filepath.ToSlash(filepath.Clean(relativePath))`
2. ok

src/linter/root.go
```diff
@@ -1556,6 +1584,25 @@ func (d *rootWalker) Report(n ir.Node, level int, checkName, msg string, args ..
...
+
+		relativePath, err := filepath.Rel(d.config.ProjectPath, filePath)
+		if err != nil {
+			d.ReportLocation(loc, level, checkName, msg, args...)
+			return
+		}
+		filePath = relativePath
+	}
+
+	var rootNode = d.config.PathRules
```
Danil42Russia, line 1600: rootNode := d.config.PathRules

src/linter/conf.go
Hidanio: 1. Why? Its okey that we can have 'bad' path or something else from the world. Function will guarantee that path would be normalize, because someone could forget about it. Why? It’s okay that we might receive a ‘bad’ path or something else from the external world. The function will guarantee that the path is normalized, because someone could forget to do it.

src/linter/root.go
Hidanio: ok, done

Hidanio: Ok, done

src/linter/conf.go
```diff
@@ -68,6 +75,54 @@ type Config struct {
 	StrictMixed bool
 }
 
+type RuleNode struct {
+	Children map[string]*RuleNode // Child nodes (subdirectories and files)
+	Enable   map[string]bool      // Rules enabled at this level
```
i582, line 80: Maybe `Enabled` and `Disabled`?

src/linter/conf.go
```diff
@@ -68,6 +75,54 @@ type Config struct {
...
+func NewRuleNode() *RuleNode {
+	return &RuleNode{
+		Children: make(map[string]*RuleNode),
+		Enable:   make(map[string]bool),
+		Disable:  make(map[string]bool),
+	}
+}
+
+type PathRuleSet struct {
+	Enable  map[string]bool // Rules that are enabled for this path
```
i582, line 93: Same here

src/linter/conf.go
Hidanio: Good catch

src/linter/conf.go
Hidanio: Good catch

PR #1244, opened 2025-01-31
linter: mask filtering rules path

Hidanio:


src/linter/root.go
```diff
@@ -1526,21 +1526,79 @@ func IsRuleEnabledForPath(root *RuleNode, filePath string, checkRule string) boo
...
+		// Now the logic is to "swallow" several directories:
+		//
+		// Until we meet an exact match in `Children` in the next step
+		// (except "*"), we can continue to "eat" directories while remaining in `starNode`.
+		//
+		// Or if we have reached the end of the path, we exit the loop.
+
+		for {
+			i++
+			if i >= len(parts) {
```
Danil42Russia, line 1579: you don't change the `parts`, why calculate its length every time?

src/linter/root.go
```diff
@@ -1526,21 +1526,79 @@ func IsRuleEnabledForPath(root *RuleNode, filePath string, checkRule string) boo
 	// Starting with global state. We have guarantee while parsing config that rule is `on` and exist
 	ruleState := true
 
-	for _, part := range parts {
```
Danil42Russia, line 1529: in go `for-range` returns the index and the value of the elements in each index. Are you sure you should have changed it to `for-loop`?

PR #1245, opened 2025-02-12
linter: null safety for functions, methods and properties for classes

Hidanio:


src/meta/metainfo.go
```diff
@@ -129,6 +129,7 @@ type FuncInfo struct {
 	Flags        FuncFlags
 	ExitFlags    int // if function has exit/die/throw, then ExitFlags will be <> 0
 	Internal     bool
+	IsVariadic   bool
```
Danil42Russia, line 132: I would combine this with `Flags`. It already has information about `IsDeprecated`, `IsPure`.
Or create separate flags for the types.

src/types/predicates.go
```diff
@@ -37,6 +37,29 @@ func IsAlias(s string) bool {
...
+	}
+	typ.Iterate(func(t string) {
+		if t == "null" || strings.Contains(t, "mixed") {
+			isNullable = true
+		}
+	})
+	return isNullable
+}
+
+func IsTypeMixed(typ Map) bool {
```
Danil42Russia, line 53: why not just `type.Contains("mixed")`? in this case, you're looking at `[]mixed` simply as `mixed`

src/linter/block.go
```diff
@@ -1027,7 +1027,202 @@ func (b *blockWalker) handleIssetDimFetch(e *ir.ArrayDimFetchExpr) {
...
 }
 
+func nullSafetyRealParamForCheck(fn meta.FuncInfo, paramIndex int, haveVariadic bool) meta.FuncParam {
+	if haveVariadic && paramIndex >= len(fn.Params)-1 {
+		return fn.Params[len(fn.Params)-1]
+	}
+	return fn.Params[paramIndex]
+}
+
+func formatSlashesFuncName(fn meta.FuncInfo) string {
```
Danil42Russia, line 1037: why? You've broken the FQN

Hidanio: Why? This is only used in the error message output.

src/types/predicates.go
Hidanio: You are right. I missed this

src/meta/metainfo.go
Hidanio: Make sense

Hidanio: Done

src/types/predicates.go
```diff
@@ -37,6 +37,23 @@ func IsAlias(s string) bool {
 	return has
 }
 
+func IsTypeNullable(typ Map) bool {
+	isNullable := false
+	if typ.m == nil {
+		return true // We consider that if the type is not inferred, then it is mixed
+	}
+	typ.Iterate(func(t string) {
+		if t == "null" || strings.Contains(t, "mixed") {
```
Danil42Russia, line 46: `t.Contains("mixed")` || `t.Contains("null")`

src/linter/root.go
```diff
@@ -501,6 +502,7 @@ func (d *rootWalker) parseFuncParams(params []ir.Node, docblockParams phpdoctype
 		return d.parseFuncArgsForCallback(params, sc, closureSolver)
 	}
 
+	isVariadic := false
```
Danil42Russia, line 505: move above where `paramHints` is declared

src/types/predicates.go
Danil42Russia: why not:

```
func IsTypeNullable(typ Map) bool {
	return typ.Contains("null") || typ.Contains("mixed")
}
```

src/linter/block.go
```diff
@@ -1027,7 +1027,202 @@ func (b *blockWalker) handleIssetDimFetch(e *ir.ArrayDimFetchExpr) {
...
+	}
+	haveVariadic := fn.Flags&meta.FuncVariadic != 0
+
+	for i, arg := range args {
+		if arg == nil {
+			continue
+		}
+
+		// If there are more arguments than declared and function is not variadic, ignore extra arguments.
+		if i > len(fn.Params)-1 && !haveVariadic {
```
Danil42Russia, line 1053: do you want to change `!haveVariadic && i > len(fn.Params)-1` to be the same everywhere?

src/linter/root.go
Hidanio: Ok

src/linter/block.go
Hidanio: Ok

src/types/predicates.go
Hidanio: Ok, but this should be: 
```
	if typ.m == nil {
		return true // We consider that if the type is not inferred, then it is mixed
	}
```

Danil42Russia (approved):

PR #1248, opened 2025-02-24
linter: expanded deprecated for calling const, property and instances of classes

Hidanio:
Now we have deprecated and deprecatedUntagged rule that could be apply for methods and functions.
I expanded it for calling const, property and instances of classes


Hidanio:
Level notice for deprecated and deprecatedUntagged

DrDet (approved):

src/linter/root.go
```diff
@@ -902,19 +907,26 @@ func (d *rootWalker) handleClassDoc(doc classPHPDocParseResult, cl *meta.ClassIn
 	}
 }
 
-func (d *rootWalker) parsePHPDocVar(doc phpdoc.Comment) (typesMap types.Map) {
-	for _, part := range doc.Parsed {
-		part, ok := part.(*phpdoc.TypeVarCommentPart)
+func (d *rootWalker) parsePHPDocVar(doc phpdoc.Comment) (typesMap types.Map, deprecationInfo meta.DeprecationInfo) {
```
Danil42Russia, line 910: `DeprecationInfo` should not be part of `parsePHPDocVar`, even the function name says that on parses `var` tag.

Hidanio: Int this case the contex of Var - variable (property). See parseConstPHPDoc, parseClassPHPDoc: in all cases we are parse everything

src/linter/block_linter.go
```diff
@@ -563,6 +556,15 @@ func (b *blockLinter) checkNew(e *ir.NewExpr) {
 		if class.IsInterface() {
 			b.report(e.Class, LevelError, "invalidNew", "Cannot instantiate interface %s", class.Name)
 		}
+
+		if class.IsDeprecated() {
+			if class.WithDeprecationNote() {
+				b.report(e, LevelNotice, "deprecated", "Try to create instance of the class %s that was marked as deprecated (%s)", class.Name, class.DeprecationInfo)
+			} else {
+				b.report(e, LevelNotice, "deprecatedUntagged", "Try to create instance %s class that was marked as deprecated", class.Name)
```
Danil42Russia, line 564: let's merge it with `deprecated`

src/linter/phpdoc_util.go
```diff
@@ -84,11 +84,11 @@ func (e *PHPDocErrors) pushLint(place PHPDocLocation, format string, args ...int
 type classPHPDocParseResult struct {
 	properties  meta.PropertiesMap
 	methods     meta.FunctionsMap
+	Deprecation meta.DeprecationInfo
```
Danil42Russia, line 87: `deprecation`

src/linter/block_linter.go
Hidanio: Done

PR #1250, opened 2025-03-11
linter: null-safety for static calls, static/non static/ property fetching

Hidanio:
[Here](https://github.com/VKCOM/noverify/pull/1245) we added checking null-safety. Here we expanded it for statics and functions


Danil42Russia (approved):
conceptually - ok

PR #1252, opened 2025-03-28
linter: better type inferring 

Hidanio:
We have some false-positives for null-safety and this pull-request should help us. I improved type inferring by true/false context and merge null-safety with more general rule: safety call (but null-safety separated)

- [x] Improve type-inferring for bool/null
- [x] Improve type-inferring for another type and checkers like is_bool, is_int, is_null and etc
- [x] Merge null-safety with more general safety-call rule

**Very important note:** i improved type inferring only for if/else conditions and context inside it. Throwing context up is the future task.

```

$x <== User|false
if ($x === false){
$y = is_object($x); <=== here will be User
 return;
}

some code....

$y = is_object($x); <== will be false; current type will be User|false

```


src/utils/utils.go
```diff
@@ -40,6 +40,15 @@ func NameNodeEquals(n ir.Node, s string) bool {
 	}
 }
 
+func ContainsKey(param string, m map[string]struct{}) bool {
```
Danil42Russia, line 43: do you really have to?

src/types/predicates.go
```diff
@@ -81,6 +85,17 @@ var trivial = map[string]bool{
...
 
+var scalar = map[string]bool{
+	"bool":   true,
+	"float":  true,
+	"int":    true,
+	"string": true,
+
+	"true":  true,
+	"false": true,
+	// "mixed": true,
```
Danil42Russia, line 96: let's clean up

src/types/lazy.go
```diff
@@ -121,8 +121,12 @@ func wrap(typ byte, byteFields []uint8, args ...string) string {
 	return string(buf)
 }
 
-func unwrap1(s string) (one string) {
-	return s[stringLenBytes+1:] // do not care about length, there is only 1 param
+func unwrap1(s string) string {
```
Danil42Russia, line 124: tell me what problem you're solving.

src/types/map.go
```diff
@@ -180,6 +188,36 @@ func (m Map) Equals(m2 Map) bool {
 	return true
 }
 
+// Union return new Map that contains all types from the current and other
+func (m Map) Union(other Map) Map {
```
Danil42Russia, line 192: what `MergeMaps` didn't triple?

src/types/map.go
Danil42Russia, line 1: everything needs tests

src/solver/solver.go
```diff
@@ -490,6 +490,53 @@ func Implements(info *meta.Info, className, interfaceName string) bool {
...
+		// Always re-read keys after modifications
+		keys := m.Keys()
+	outer:
+		for i := 0; i < len(keys); i++ {
+			for j := i + 1; j < len(keys); j++ {
+				key1 := keys[i]
+				key2 := keys[j]
+				if IsMoreSpecific(metaInfo, key1, key2) {
+					m = m.Erase(key2)
+					changed = true
```
Danil42Russia, line 527: why not just `return m`?

src/types/lazy.go
Hidanio: Check 'IsBoolean' in `map.go`. I wanna check can we unwap bool element.

src/types/map.go
Hidanio: Because we wanna union current with another vs Create new a lot of maps. Its ok to separate it

src/solver/solver.go
Hidanio: Yor are right

src/utils/utils.go
Hidanio: Removed

src/types/predicates.go
Hidanio: Removed

src/types/lazy.go
```diff
@@ -121,8 +121,12 @@ func wrap(typ byte, byteFields []uint8, args ...string) string {
...
+	if len(s) > stringLenBytes+1 {
+		return s[stringLenBytes+1:]
+	}
+
+	return ""
```
Danil42Russia, line 129: I don't like the fact that it returns an empty string.
If this is happening, we probably have a problem here

Hidanio: Previously, in case of incorrect use there was a panic at the unwrap level. Now panic or an error will occur at the logic level. There should be no problems

Danil42Russia: Most likely we are looking at types before they were wrapped

src/types/map.go
Danil42Russia: I don't get it. We've created a new map here or there

src/types/lazy.go
Hidanio: Yes, but in my case I want to try to unwrap an array element. In any case, this logic will stop causing panic at the unwraping level, and will cause an error above. It will be immediately clear, there are no hidden traps here

src/types/map.go
Hidanio: Union is logically tied to the current instance and merges a specific map with another, while MergeMaps can be used to merge any map

Danil42Russia (approved):

PR #1256, opened 2025-06-23
general: git full diff bug fix

Hidanio:
The problem was that second part did with not the same instructions as the first


PetrShumilov (approved):
LGTM

